PR #6, opened 2020-10-13
Update __init__.py

kevingoh:
Found a bug in RSSFeedStoryProvider that if the passed limit is more than the returned feed entries itself an exception will be thrown.


j6k4m8:
Awesome, thank you so much!!

PR #29, opened 2020-12-08
alpine-based Dockerfile

lsmoura:
Would the project be interested in an alpine-based Docker image?

It takes a while to complete because it needs to compile stuff but the end result is an image with less than half the size (395MB vs 1.09GB)

The usage would be something like:

`docker run --rm -v $(pwd):/data j6k4m8/goosepaper:alpine -c /data/example-config.json -o /data/output.pdf`

It also uses the current files on the filesystem instead of downloading the files from github.

The build is made with reusability and size in mind:
* during development, if you're building the image over and over again, only the last few steps are going to be executed
* no build dependencies are ever committed to the docker image, keeping the size small
* whenever building for new versions, the dependencies part will most likely remain unchanged, so the first few steps will be reused

If you'd like to use the new dockerfile:
* Feel free to change the maintainer tag if you like
* Feel free to test it using the image that I've build on dockerhub: https://hub.docker.com/r/lsmoura/goosepaper (using `docker pull lsmoura/goosepaper:0.3.0`)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.github
+.git
+.DS_Store
+docs
+Dockerfile
+*.pdf
+*.json
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -1,0 +1,37 @@
+FROM alpine:3.12
+
+LABEL maintainer="Sergio Moura <sergio@moura.ca>"
+
+RUN apk --update --no-cache add cairo libffi libjpeg libstdc++ libxml2 libxslt pango \
+py3-aiohttp py3-cffi py3-feedparser py3-gobject3 py3-html5lib py3-lxml py3-multidict \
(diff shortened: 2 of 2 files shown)


j6k4m8:
Sorry for the delay in getting back to you @lsmoura! It's been high on my list to bring the size of this image down a bit, thank you so much for this contribution! Would be happy to leave you as maintainer if you are willing, but otherwise I'm happy to add it to the automatic rebuild when someone pushes to this repo!

lsmoura:
I'd be very happy to be listed as a maintainer!

I'm glad that you like having small images. It takes longer to build, sure, but in the end everyone benefits.

j6k4m8:
Awesome. With your approval, I'll merge this in!

lsmoura:
Approved! 🥇

PR #36, opened 2021-03-21
Allow to replace documents on upload

traysh:
Good to go!


j6k4m8:
Back at it again with the slick feature updates!! Nice!

No pressure, but changelog update is super appreciated so you get the street cred :)

traysh:
fixed

j6k4m8:
@traysh absolutely killing it :) Merging!!

PR #38, opened 2021-03-24
Allow upload to a folder

traysh:
This makes the repo up to date to the local changes I had (but organized in commits :))

for reference, I have multiple JSON configs and upload all of them to a "News" folder in the Remarkable Cloud. The command I use is:

```
for i in *.json
do
    if [[ "$i" == example-config.json ]]
        then continue
    fi
    (goosepaper --config "$i" --upload) &
done
wait
rm *.pdf
```

in one line: `for i in *.json; do if [[ "$i" == example-config.json ]]; then continue; fi; (goosepaper --config "$i" --upload) &; done; wait; rm *.pdf`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
     -   Parallelize fetching RSS stories - contribution by @traysh
     -   Add flag to allow replacing documents in remarkable cloud - contribution by @traysh
     -   Add class to resolve options both from command line and config file - contribution by @traysh
+    -   Allow uploading to a folder in remarkable cloud
 
 ### **0.3.0** (November 27 2020)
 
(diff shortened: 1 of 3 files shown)


j6k4m8:
Cool! Does this work with nested folders?

`/news/news_about_geese/Todays_Goose.pdf`?

traysh:
hm... no, it does not, it will create the document in no folder at all.
I won't handle nested folders for now, but I'll check if the folder was successfully created. I'll update this branch as soon as I finish doing it

traysh:
@j6k4m8 lots of checks. Warns nested folders are not supported if the user tries to do that.

j6k4m8 (approved):
This rocks! Thank you a ton for the extra checks —  I think it's super helpful to get that telemetry or fail-info back to the user!

PR #39, opened 2021-04-05
Add option to only fetch new RSS stories

traysh:
@j6k4m8 what do you think about this?


j6k4m8:
I like this! I'm curious why you're now explicitly passing `stories` to the html function? Seems redundant with the (previous) first line, no?

Issue #44, opened 2021-05-10
Global preferences


sedennial:
I think there should be a global preferences config file.

j6k4m8:
Love this idea. Maybe something that extends the current goosepaper config, so that your global config looks like a goosepaper config, but a specified config file overrides those values? 

I propose that because I think it'd be convenient to have a single json spec that we have to parse...

sedennial:
I like that idea too which is kinda what I was thinking. For example I'd like to have my default font size at 12, but on a couple of RSS feeds that are very verbose I'd like those to be at 10.

Plus I don't have a single big config file since I run all mine as separate cronjobs. Just a personal idiosyncrasy.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

sedennial:
I'm working on this right now (mostly done) with a hard coded order of precedence as well as add ability for certain new options to be changed on a per story basis (like you can with font-size). The order of priority (from least to highest) is

1. $HOME/.goosepaper.json
2. $(CWD)/goosepaper.json (with CWD being the cwd of the shell at the time it called goosepaper, not the goosepaper directory. So this can be specified in either cron or windows task scheduler/.lnk file. Allows for you to have separate directory + config file (if you wanted) for each story job and contain every operational value in the config. No longer HAVE to specify ANY command line argument if you don't want to.
3. --config <configfile>.json
4. option-embedded-in-story-config
5. --<commandline-option>

sedennial:
Oh, and I'm also adding a default 'goosepaper.json' which contains all (existing and new/pending) options with all options set to 'safe' mode where 'safe' means keep the functionality as it exists in 0.3.1 (0.3.7?) so as to not break existing workflow processes or expected behavior.

This means by default some of the new sanity checks options (when finalized) will be disabled.

PR #48, opened 2021-05-14
Multiple fixes, feature adds, cleanup. Also partially completes issue 44…

sedennial:
… #44

Fixes:
	Renamed upload function to do_upload as we use the option 'upload' as a boolean elsewhere. Confusing.
	Fixed a missing default for 'upload' in multiparser class that wasn't allowing 'upload' to be used as a config file option.

Cleanup/Housekeeping:
	Changed some errors handling from 'raise error' to simply print error message and exit with a exit code of one (1). Some errors such as invalid extension or file not found don't need the traceback they just need to output notification to the user and set the shell exit code as they are user errors, not code errors.
	Moved some variables to upload since they are only needed there.
	Moved the story builder inside a conditional to add ability to skip story generation via command line switch (--nostory), or individual story option. Will allow you to temporarily turn off a particular story without removing it from your config or leave your workflow in place without actually deleting your tasks/job. Also handy for testing upload functionality from the command line.

New features:

  * Added --nostory option. Allows you to skip all the story generation and process an upload only. In this case the 'output' file is used as the file to upload directly.
  * Added '--noreplace' and '--noupload' as command line only options to allow overriding config file settings (for testing (or other uses ?)) without having to constantly edit config files.
  * Added 'upload' specified in config or global
  * Added 'output' can be specified in config, global, or command line.
  * Added sanity checking if you provide more than one 'output'.
  * Created default config files
  * Added replace to default options as 'False'
  * Added cleanup option (delete target goosepaper after upload success)
  * Added formatter class to multiparser to allow newlines (optionally) used in help for command line options. Makes for cleaner reading on some help text (IMO). Mostly mine since my default setting is --verbose --verbose. :)
  * Added a per story 'skip' option (similar to nostory global option but per story source) which will allow you to skip processing a particular story provider without having to remove your entire story config. Useful if the source is having issues but you don't want to remove your entire entry from that config.
  * Added an entry to examples to show use of 'skip'
  * Added checks to make sure that either at least one default config file exists or you specified one on the command line.
  * Added global config preference files in order of precedence. Command line options superseded all config file settings.
      ~/.goosepaper.json (HOME environment from os)
      ./goosepaper.json (current working directory when goosepaper called)
      --config <configfile.json> (specified as argument)
  * Added '--showconfig' command line switch to allow both user and dev debugging of config file and command line options since there are four ways to pass settings and options. Shows each file in order parsed as well as the final config
  * Added goosepaper.json containing all the (current) operational variables set to operate with the least impact to any existing workflow process (i.e. no change to current behavior as of 0.3.1).

### DIFF
--- a/example-config.json
+++ b/example-config.json
@@ -30,6 +30,14 @@
         {
             "provider": "reddit",
             "config": { "subreddit": "todayilearned" }
-        }
+        },
+	{
+	    "provider": "rss",
+	    "config": {
+		"rss_path": "https://krebsonsecurity.com/feed/",
+		"limit": 1,
+		"skip": true
+	    }
+	}
     ]
 }
--- a/goosepaper/multiparser.py
+++ b/goosepaper/multiparser.py
@@ -1,6 +1,14 @@
 import argparse
+import os
+from os import getenv
 
 from goosepaper.util import load_config_file
+
+class NewLineFormatter(argparse.HelpFormatter):
+    def _split_lines(self, text, width):
+        if text.startswith('||'):
+            return text[2:].splitlines()
+        return argparse.HelpFormatter._split_lines(self, text, width)
 
 class MultiParser:
     def __init__(self):
@@ -12,13 +20,15 @@ class MultiParser:
 
         self.parser = argparse.ArgumentParser(
             prog="goosepaper",
-            description="Goosepaper generates and delivers a daily newspaper in PDF format."
+            description="Goosepaper generates and delivers a daily newspaper in PDF format.",
+            formatter_class=NewLineFormatter
         )
+
         self.parser.add_argument(
             "-c",
             "--config",
             required=False,
-            default=None,
+            default="",
             help="The json file to use to generate this paper.",
         )
         self.parser.add_argument(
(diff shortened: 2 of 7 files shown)


goosepaper/util.py
```diff
@@ -74,7 +74,7 @@ def construct_story_providers_from_config_dict(config: dict):
         return []
 
     stories = []
-    import inspect
```
sedennial: Accidentally left this in from debug.

example-config.json
```diff
@@ -30,6 +30,14 @@
         {
             "provider": "reddit",
             "config": { "subreddit": "todayilearned" }
-        }
+        },
```
j6k4m8, line 33: Is there a reason you added this to the example file? (I happen to read KoS plenty, just wondering since it adds runtime to the example.)

goosepaper/multiparser.py
```diff
@@ -27,35 +37,150 @@ def __init__(self):
...
         self.parser.add_argument(
             "-u",
             "--upload",
             action="store_true",
             required=False,
+            default=None,
             help="Whether to upload the file to your remarkable using rmapy.",
         )
+        self.parser.add_argument(
+            "--noupload",
```
j6k4m8, line 55: I think `--upload/--noupload` should be [mutually exclusive](https://docs.python.org/3/library/argparse.html#mutual-exclusion) to avoid the weirdness of a command like `goosepaper --upload --noupload myfile.pdf`

goosepaper/multiparser.py
```diff
@@ -27,35 +37,150 @@ def __init__(self):
...
+            required=False,
+            default=None,
+            help="Overrides any other 'upload: true' arguments from config files or command line. Useful for testing configs or story generation without having to edit config files."
+        )
+        self.parser.add_argument(
+            "--showconfig",
+            action="store_true",
+            required=False,
+            default=None,
+            help="Print out all config files and command line options in order loaded and the final config to help finding conflicting options. Needed since there are now four possible ways to pass options."
```
j6k4m8, line 66: omg i love this idea — super helpful now that we are starting to cascade options from a global config :)

goosepaper/multiparser.py
```diff
@@ -27,35 +37,150 @@ def __init__(self):
...
+        )
+        self.parser.add_argument(
+            "-n",
+            "--nostory",
+            required=False,
+            default=False,
+            action="store_true",
+            help="||Skip story creation. Combined with \"--upload\" can be used to\nupload a preexisting output file.\n\n** NOTE ** If used without \"--upload\" goosepaper will run but\nperform no action.",
```
j6k4m8, line 74: Not totally convinced this option makes sense here — more on that in the review notes!

goosepaper/multiparser.py
```diff
@@ -27,35 +37,150 @@ def __init__(self):
...
+            )
+        self.parser.add_argument(
+            "--cleanup",
+            required=False,
+            default=None,
+            action="store_true",
+            help="Delete the output file after upload."
+        )
+        self.parser.add_argument(
+            "--nocase",
```
j6k4m8, line 98: Maybe something like `--ignorefilenamecase`? Trying to make it clear that "case" here refers to the _filename_... (This context might not be obvious to a user!)

goosepaper/multiparser.py
```diff
@@ -27,35 +37,150 @@ def __init__(self):
...
+            default=None,
+            action="store_true",
+            help="** COMING SOON: Enables case insensitivity on upload. By default folders \"My Goosepapers\" and \"my goosepapers\" are NOT the same folder. Enabling this feature means that they would be considered the same. No effect unless \"-u\" is specified."
+        )
+        self.parser.add_argument(
+            "--strictlysane",
+            required=False,
+            default=None,
+            action="store_true",
+            help="** COMING SOON: Force strict sanity checking, implies \"--no-case\". Fail on ANY duplicate folder or document names in the root or target upload folders. No effect unless \"-u\" is specified."
```
j6k4m8, line 109: I think this is currently `--nocase`, right?
```suggestion
            help="** COMING SOON: Force strict sanity checking, implies \"--nocase\". Fail on ANY duplicate folder or document names in the root or target upload folders. No effect unless \"-u\" is specified."
```

goosepaper/multiparser.py
```diff
@@ -27,35 +37,150 @@ def __init__(self):
...
+        )
+        self.parser.add_argument(
+            "--nocase",
+            required=False,
+            default=None,
+            action="store_true",
+            help="** COMING SOON: Enables case insensitivity on upload. By default folders \"My Goosepapers\" and \"my goosepapers\" are NOT the same folder. Enabling this feature means that they would be considered the same. No effect unless \"-u\" is specified."
+        )
+        self.parser.add_argument(
+            "--strictlysane",
```
j6k4m8, line 105: This made me laugh, but it definitely doesn't give me any idea what this flag does as an end-user. Maybe something like `--avoid-filename-collisions`...? I hate to be a funsucker here but my suspicion is that goosepaper cli commands are about to get way more verbose with all these (awesome!) new options, so I think it makes sense to make each option as clear as possible.

goosepaper/multiparser.py
```diff
@@ -27,35 +37,150 @@ def __init__(self):
...
-            ) from e
+        # These are in order of precedence, low to highest
+        #  1. Home directory global configs
+        #  2. Local directory from which goosepaper is called
+        #  3. Specified on the command line.
+        
+        defaultconfigs = [ os.getenv('HOME') + "/.goosepaper.json", './goosepaper.json', self.args.config ]
+        foundconfig = False
+        self.config = {}
+        outputcount=0
```
j6k4m8, line 121: I think this isn't styled according to the `black` styleguides. I don't have that linting set up automatically anywhere, but it might be useful to run [`black`](https://github.com/psf/black) on this branch to quickly clean up operator spacing things like this :)

goosepaper/multiparser.py
```diff
@@ -27,35 +37,150 @@ def __init__(self):
...
+            pp.pprint (self.args)
+
+        # If passed a config file on the command line, assume it's important so fail
+        # if not readable.
+
+        if self.args.config:
+            try:
+                valid = load_config_file(self.args.config)
+            except FileNotFoundError:
+                print ("Honk! Honk! Somebody stole my egg! Couldn't find config file ({0}) specified on the command line. Aborting migration.".format(self.args.config))
```
j6k4m8, line 138: 🥺 🥚

j6k4m8 (commented):
You've got some really amazing changes in here, thank you so much!! I can't wait to start using some of these :)

I do hesitate a little bit to have upload-only code in here (i.e. a code path that JUST uploads and has nothing to do with paper generation). I worry that we're starting to mix code purposes a little, and maybe upload-specific workflows should use something like [`rmapy`](https://github.com/j6k4m8/rmapy/) directly.

I can be easily convinced about this, though :) 

I'd definitely run `black` for code-linting on this branch to make our life easier on styling fixes (I've been meaning to add a test check in GitHub actions to make sure everything passes the `black` styleguide). Otherwise, curious to hear your comments on the notes above!

PS: Sorry for taking a while to review this — it's been a wild few days over here!

example-config.json
sedennial: I didn't add that. It was already there.

goosepaper/multiparser.py
sedennial: That would require some significant mods to the multiparser to determine if upload is present on the command line only or in the config file. My thought process also was the allowing '--noupload' combined with --showconfig on the command line would allow easier testing/validation of configs without having to tweak config files back and forth. 

Poor logic on my part?

sedennial:
> I do hesitate a little bit to have upload-only code in here (i.e. a code path that JUST uploads and has nothing to do with paper generation). I worry that we're starting to mix code purposes a little, and maybe upload-specific workflows should use something like [`rmapy`](https://github.com/j6k4m8/rmapy/) directly.

I kinda thought about that too, but one reason I did it was if you have a lot of goosepapers (I have five separate ones which take a while to run) the user could have them created, and then have a different upload process entirely if they wanted. Which is how I was using it initially.  While right now the users probably have the ability to use rmapy to do standalone uploads, it would be nice for less programmatic users to be able to use GP to handle that too.  Right now I'm also using pandoc to produce some daily outputs from non-RSS sites (https://isc.sans.org) for example and I'm using GP to handle the upload.

Also in line with the possibility of sending update, documentation, errors, etc. to the RM, this gives GP the ability to compile multiple logs from multiple runs, convert them, and then upload them.

All of that said I'm not going to dig in my heels if you want to take it out. I'll just keep it in my private branch. :)

goosepaper/multiparser.py
sedennial: Hmm....originally I intended it to be both folder and filename...but now that you say that, perhaps doing --ignorefilenamecase and --ignorefoldernamecase would be better? I don't know WHY you'd want to ignore one but not the other, but hey.... :)

And '--strictsanitychecking' would imply both case options set to true.

goosepaper/multiparser.py
sedennial: I still like it. It makes testing config parse order easier when combined with '--showconfig'. I HAD to put it in in order to make validating why options were conflicting given four different mechanisms for consuming options.
:)

goosepaper/multiparser.py
sedennial: No it wasn't the same. '--strictlysane' would also force checks for duplicate folders and files in the root (or target upload) directory and abort. But if we use separate case options (like above) it could honor those options and only terminate on case policy violations.  Even with 'nocase' options it is still possible to have case identical duplication of file and folder names.

goosepaper/multiparser.py
sedennial: I think the strictlysane and case options need to be rethought. I'm going to pull them out for now.

goosepaper/multiparser.py
sedennial: Fixed.

example-config.json
sedennial: Let me clarify..if you run the example against my PR it isn't processed. If you run in against the current master it would be.

j6k4m8:
@sedennial — I had some stylistic nitpicks which is why it's taken me so long to get back to this — and then I realized that if I really was going to be such a stickler, there were plenty of places to fix my own code first :) Very sorry for the lag here.

My main priority is making sure that this is backwards-compatible with prior code (i.e., someone can update their goosepaper install and keep using their scripts as-is). So far I think this is true, but insight welcome! Otherwise, I think this is ready to merge, right?

j6k4m8 (approved):

PR #55, opened 2021-05-29
Allow optional single-threaded RSS fetching

### LINKED ISSUE #53: RSS feed story providers not working
While trying to write documentation for the storyproviders I noticed the RSS providers aren't working.  
I have tried different feeds, but I'm not getting any error-messages or any stories in my pdf-export.  
Anyone else having this issue? Anyone know what could cause this?

j6k4m8:
This enables optional single-threaded fetching of RSS feeds with `"parallel": false` in the config.

This potentially fixes #53 — separately, the default RSS example config was broken! So this PR also resolves that.

@kwillno — does this fix the issue you were encountering?

### DIFF
--- a/docs/Customizing.md
+++ b/docs/Customizing.md
@@ -177,7 +177,7 @@ The parameters are as follows:
 ```json
 	"rss_path"	:	"https://feeds.npr.org/1001/rss.xml" (str 	- RSS feed you want to see results from.)
 	"limit"		:	20 				(int 	- Amount of reddit headlines to provide.)
-	"paralell"	:	true 			(bool 	- Whether to use multiprocessing to fetch in parallel)
+	"parallel"	:	true 			(bool 	- Whether to use multiprocessing to fetch in parallel)
 ```
 
 ### <a name="Twitter">Twitter</a>
--- a/docs/Customizing.md
+++ b/docs/Customizing.md
@@ -27,7 +27,7 @@ As an example we give the config delivered as an example `example-config.json`:
         {
             "provider": "rss",
             "config": {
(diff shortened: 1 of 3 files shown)


kwillno:
Yes, seems to work now. Couldn't get the parallelization to work though, but that might only be my setup. Took longer to render the same rss-source with `"parallell" : true` than with it disabled.  
Also noticed you misspelled `parallel` as `"paralell"` in the docs, seems to only work with `"parallel"`. Checked both of them in my config-file.  
Nice work!

docs/Customizing.md
```diff
@@ -94,123 +98,136 @@ Stories and storyproviders are given in the config-file using the `"stories"`-ke
...
 
 The parameters are as follows:
-```json	
-	"rss_path"	:	"https://www.npr.org/feed/"	(str 	- RSS feed you want to see results from.) 
-	"limit"		:	20 				(int 	- Amount of reddit headlines to provide.) 
+
+```json
+	"rss_path"	:	"https://feeds.npr.org/1001/rss.xml" (str 	- RSS feed you want to see results from.)
+	"limit"		:	20 				(int 	- Amount of reddit headlines to provide.)
+	"paralell"	:	true 			(bool 	- Whether to use multiprocessing to fetch in parallel)
```
j6k4m8, line 180:
```suggestion
	"parallel"	:	true 			(bool 	- Whether to use multiprocessing to fetch in parallel)
```

j6k4m8:
Ah shoot, good catch! I guess that's what I get for committing code late at night. Fixed the typo!

PR #73, opened 2022-08-14
Styles-as-plugin

zerone01y:
The styles were hard-coded in the package; reading them from a styles folder will make goosepaper much easier to customize.

These codes basically do one thing: look for a user-defined style in styles folder, and read it on-the-fly. It keeps one hard-coded style (originally FifthAvenue; now it will only be used when no /wrong style is configured).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ docs
 Dockerfile
 *.pdf
 *.json
+*.yml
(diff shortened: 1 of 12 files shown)


j6k4m8:
This is an AWESOME update, and really cleans up some code I was unhappy about in the original repository :) Could you separate out the changes that are unrelated to style (e.g., the RSS storyprovider changes) so we can merge that in separately? I want to isolate changes in PRs if possible. 

Thank you a bunch for contributing, this is super!!

PR #74, opened 2022-08-17
Styles as plugin

zerone01y:
The styles were hard-coded in the package; reading them from a styles folder will make goosepaper much easier to customize.

These codes basically do one thing: look for a user-defined style in styles folder, and read it on-the-fly. It keeps one hard-coded style for robustness (originally FifthAvenue; now it will only be used when no/wrong style is configured).

---

I closed the previous PR as it was a bit messy. Glad you like this!

I have a small piece [here](https://github.com/zerone01y/goosepaper/blob/master/README.md#edit-your-own-styles) for instructions but it is not included in this PR, as I later found out you have a doc for customising (btw we should put a link to this doc in readme). But just let me know if you want to add the instructions.

### DIFF
--- a/goosepaper/goosepaper.py
+++ b/goosepaper/goosepaper.py
@@ -1,3 +1,4 @@
+import pathlib
 from typing import List, Optional, Type, Union
 import datetime
 import io
@@ -169,7 +170,10 @@ class Goosepaper:
         style_obj = _get_style(style)
         html = self.to_html()
         h = HTML(string=html)
-        c = CSS(string=style_obj.get_css(font_size),font_config=font_config)
+        base_url = str(pathlib.Path.cwd())
+        c = CSS(string=style_obj.get_css(font_size),
+            font_config=font_config,
+            base_url=base_url)
         # Check if the file is a filepath (str):
         if isinstance(filename, str):
             h.write_pdf(
--- a/goosepaper/goosepaper.py
+++ b/goosepaper/goosepaper.py
@@ -144,7 +144,7 @@ class Goosepaper:
     def to_pdf(
         self,
         filename: Union[str, io.BytesIO],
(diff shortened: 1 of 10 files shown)


j6k4m8:
This is super awesome — can you run [`black`](https://github.com/psf/black) on the new code so that it matches the existing Goosepaper style? 

I think this is ready to go, though we may need to add these files to the manifest so that they're included when we pip-install. I'm not at a computer right now but I think that pip-installing goosepaper without adding these non-python files to a [MANIFEST.in file](https://packaging.python.org/en/latest/guides/using-manifest-in/) will break.

So amazing, thank you for this great change :)

zerone01y:
Thanks for the great suggestion! I'll look into them.

The code requires to mount the styles folder and a goosepaper.json file, or to run the command in a directory containing these files after `pip install`. This will make it easier to customize, and that's why I put the styles folder outside the `goosepaper` package. With these aims, I feel it is fine to exclude the non-python files. After all, a default CSS script is kept in 'styles.py' for people who don't want to customize it. 

Let me know what you are thinking of the issue.

j6k4m8:
This is neat — I think it's fair game to include the styles directory in the main repository though! I'm supportive of allowing users to register their _own_ styles, but I think it's fair for us to have a few "built-in" themes. Thoughts on that?

zerone01y:
Please check again!

zerone01y:
I noticed some irrelevant code was incorrectly included in the commits and broke the test. Already removed it, please test again.

j6k4m8 (approved):
Sorry for the delay, I've been going through the quals process for grad school and have been quite busy. This looks great to me, ready to merge @zerone01y?

zerone01y:
Definitely!

j6k4m8:
Thank you @zerone01y!!

Issue #75, opened 2023-02-04
pdf page size


martinetd:
Hello!

I've been playing with goosepaper last week, it's great -- fiddling a bit with it to read less on computer in front of kids and it looks like I'll be able to work some nice workflows.

One problem I've run into is page size; I'm not sure how it used to render on older versions but with the latest 3.0 update I find the remarkable is pretty bad at handling A4 pdf (default paper size) -- there's often some text left over at the bottom, requiring to scroll, which is horrible UX.

I'm sure we could muck up with margins, but it's just as simple to get page size right: just add `size: x y;` to the style in `@page` you use and it'll come out alright (for the remarkable2, the screen size is supposed to be 188mm by 246mm, with a resolution of 1404 by 1872px (ratio doesn't match, eh).
Doing full 1404x1872 px makes the font too small, but I've had good results with `size: 184.5 246mm;` and `size: 702pt 936pt;` (fits more text, mm gets converted to something in the 500s px)

I've also made the middle colon slightly wider with `column-gap: 2rem;` in the last `.row` of the FifhtAvenue stylesheet.css.

Papers with these changes look fairly good for me (also made the header smaller as so much blank space feels weird without weather; it's probably safe to ignore that part; margins are also mostly personal taste and you might be better with the original values this is just an example you'll probably want to fiddle with hence not a PR)

```diff
diff --git a/styles/FifthAvenue/stylesheet.css b/styles/FifthAvenue/stylesheet.css
index 181d56ff1505..87b400917954 100644
--- a/styles/FifthAvenue/stylesheet.css
+++ b/styles/FifthAvenue/stylesheet.css
@@ -1,15 +1,16 @@
 @page {
-    margin-top: 0.5in;
-    margin-right: 0.2in;
-    margin-left: 0.65in;
-    margin-bottom: 0.2in;
+    size: 702pt 936pt;
+    margin-top: 30mm;
+    margin-bottom: 20mm;
+    margin-left: 20mm;
+    margin-right: 10mm;
 }
 body {
     font-family: "Open Sans";
 }
 .header {
-    padding: 1em;
-    height: 10em;
+    padding: -6em;
+    height: 6em;
 }
 .header div {
     float: left;
@@ -93,4 +94,5 @@ figure>span {
 }
 .row {
     column-count: 2;
-}
\ No newline at end of file
+    column-gap: 2rem;
+}
```

j6k4m8:
[...]

PR #77, opened 2023-03-24
styles: add page size, review margins

martinetd:
weasy print's css supports the @page 'size' attribute, allowing us to specify a pdf size different from the default A4 size.

The remarkable is 1872 x 1404 pixels, but using that makes fonts too small without further adjusments so setting a page size of half that is easier to use.

Having a proper page size, we can re-adjust the margins to better fit the new format. Also made the space between columns in FifthAvenue slightly bigger as a matter of taste.

Fixes: #75

------
Sorry for the delay, here's the PR. Since I brought it up in the issue I've also done a similar modification to Academy as some documents render better with a single column.

The margins are up to discussion (sorry for switching to metrics, I couldn't understand inches if my life depends on it -- might want to switch to something like em for compromise if you're similar the other way around :P) ; I honestly didn't spend very long fiddling with them.
I'd attach a sample document with that but github doesn't seem to allow attachments here, so I guess we'll do without...

### DIFF
--- a/styles/Academy/stylesheet.css
+++ b/styles/Academy/stylesheet.css
@@ -1,8 +1,9 @@
 @page {
-    margin-top: 0.5in;
-    margin-right: 0.2in;
-    margin-left: 0.65in;
-    margin-bottom: 0.2in;
+    size: 702pt 936pt;
+    margin-top: 30mm;
+    margin-bottom: 20mm;
+    margin-left: 20mm;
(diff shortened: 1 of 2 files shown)


j6k4m8:
This is so awesome @martinetd!! Thank you so much for contributing! Pending these status checks, I'll merge it in :D 🎉

PR #112, opened 2026-05-29
fix: correct rss feed handling for large feed counts

duncanam:
# Summary
Hello! I've got another suggestion. Let me know what you think- it's much broader than my last PR, so probably needs a bit more careful consideration.

Feed sizes greater than the feed object's default limit would still revert to the feed object's default limit due to the `min` call. By removing the limit from `get_stories`, it simplifies the code and deletes the requirement to fix this logic at all. I discovered this by passing an RSS feed limit of 15, which when fed into `get_stories`, was compared against the feed object's default stateful limit of 5, which took precedent in the `min` call to determine the limit. 

## Fixes
One option is to change from duplicating the default int limits in the `get_stories` method to `Option[int] = None`, and then add a conditional check when `None` to default to using `self.limit`. However, this felt like a code smell so instead I opted to just delete the limit argument entirely since it was only used in testing. 

## Verification
As seen here, if a limit of 15 is supplied, the RSS feed correctly collects them instead of limiting to 5. 

Source line:

```
        {
            "type": "rss",
            "url": "https://feeds.arstechnica.com/arstechnica/index",
            "limit": 15,
            "byline": "first",
            "body_source": "article"
        }
```

Resulting document:

image

### DIFF
--- a/goosepaper/storyprovider/test_bluesky.py
+++ b/goosepaper/storyprovider/test_bluesky.py
@@ -67,8 +67,8 @@ def test_bluesky_provider_uses_public_author_feed(monkeypatch):
 
     monkeypatch.setattr(bluesky.requests, "get", fake_get)
 
-    provider = bluesky.BlueskyStoryProvider("jordan.matelsky.com", limit=4)
-    stories = provider.get_stories(limit=2)
+    provider = bluesky.BlueskyStoryProvider("jordan.matelsky.com", limit=2)
+    stories = provider.get_stories()
 
     assert seen["url"] == "https://public.api.bsky.app/xrpc/app.bsky.feed.getAuthorFeed"
     assert seen["params"] == {
@@ -99,7 +99,7 @@ def test_bluesky_provider_can_exclude_replies(monkeypatch):
         "jordan.matelsky.com",
         include_replies=False,
     )
-    stories = provider.get_stories(limit=2)
+    stories = provider.get_stories()
 
     assert seen["params"]["filter"] == "posts_no_replies"
     assert len(stories) == 1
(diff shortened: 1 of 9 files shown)


goosepaper/storyprovider/test_bluesky.py
```diff
@@ -67,8 +67,8 @@ def fake_get(url, *, params, headers, timeout):
 
     monkeypatch.setattr(bluesky.requests, "get", fake_get)
 
-    provider = bluesky.BlueskyStoryProvider("jordan.matelsky.com", limit=4)
-    stories = provider.get_stories(limit=2)
+    provider = bluesky.BlueskyStoryProvider("jordan.matelsky.com", limit=2)
```
duncanam, line 70: Had to set this to pass the assert below. I don't *think* this has negative implications but I could be missing something.

goosepaper/storyprovider/test_bluesky.py
```diff
@@ -99,7 +99,7 @@ def fake_get(url, *, params, headers, timeout):
         "jordan.matelsky.com",
         include_replies=False,
     )
-    stories = provider.get_stories(limit=2)
```
duncanam, line 102: The test objects now all use their default values; hopefully this is okay.

duncanam:
@j6k4m8 let me know if you have any thoughts here!

j6k4m8:
Ooh @duncanam sorry for the delay, this slipped past me! This makes perfect sense to me!

j6k4m8 (approved):
This looks great, @duncanam is it ok for me to merge? :)

PR #120, opened 2026-07-29
fix: wrap long code blocks instead of letting them overflow the column

Smengerl:
## What

`<pre>`'s default `white-space: pre` never wraps long lines. In a browser that just means a horizontal scrollbar - there's no scrolling on a printed PDF page, so a code block wider than its column silently overflowed into whatever was rendered next to it. In a multi-column layout that's the next column's text; observed with a JS code snippet from a real RSS-sourced dev-blog article, in a 2-column newspaper, producing an unreadable, overlapping mess.

`<pre>` was also completely unstyled before this - no visual distinction from regular body text, no cue that it was a code block at all.

## Reproduction

Rendered a minimal story through Goosepaper's real pipeline (`FifthAvenue`
style, `paper_pro` profile, `2col` layout) with a `<pre><code>` block
containing one long, unbroken line (a realistic JS one-liner, the kind a
dev-blog code sample often has).

**Before (master):** the long line runs straight off the right edge of the
page instead of wrapping - worse than just bleeding into the next column,
it's outright clipped/lost past the page boundary.

**After (this branch):** the block wraps cleanly within its column, with a
light background/border so it actually reads as a code block.

*(Screenshots attached to this PR - same repro script, only the branch differs.)*
image-1785355114858
image-1785355122034

## Fix

- `pre, code { white-space: pre-wrap; overflow-wrap: break-word; word-break: break-word; }`
  - `pre-wrap` keeps existing line breaks/indentation but allows wrapping;
  `overflow-wrap`/`word-break` as a second line of defense for a single
  unbroken token (a long URL, hash, or identifier) still wider than the
  column on its own.
- `pre { max-width: 100%; ... background/border/font-size/line-height }` -
  gives code blocks an actual visual identity instead of blending into body
  text.
- `:not(pre) > code { ... }` - separate, lighter styling for genuinely
  inline code (`<code>` not inside a `<pre>`), distinct from block code.

## Testing

- New test `test_code_blocks_wrap_instead_of_overflowing_the_column` in `test_styles.py`, asserting the generated CSS contains the wrap/overflow rules.
- Full existing test suite passes (73 passed).
- Manually reproduced the overflow on `master` and confirmed it's fixed on this branch (see screenshots).

## Scope

Pure CSS addition to the shared base stylesheet - no theme currently re-declares `pre`/`code`, so this applies uniformly across every style without any cascade conflicts. `fix/svg-overflow` (same function, non-overlapping rules) is already merged as #119.

**Overlap note:** #126 (`fix/hide-interactive-buttons-in-print`) also inserts a rule block into `_base_print_css` right next to this one. The two rule sets don't interact, but merging both will hit a textual conflict at the shared insertion point (git can't tell the two independent insertions apart) - trivial to resolve by keeping both blocks, no semantic decision needed either way.

### DIFF
--- a/goosepaper/test_styles.py
+++ b/goosepaper/test_styles.py
@@ -1,0 +1,11 @@
+from .styles import Style
+
+
+def test_code_blocks_wrap_instead_of_overflowing_the_column():
+    """A <pre> block wider than its column must wrap, not overflow into whatever renders next to
+    it - see the multi-column overlap this was fixed for."""
+    css = Style("FifthAvenue").get_css(page_profile="paper_pro", layout="2col")
+
+    assert "white-space: pre-wrap" in css
+    assert "overflow-wrap: break-word" in css
+    assert "max-width: 100%" in css
--- a/goosepaper/test_styles.py
+++ b/goosepaper/test_styles.py
@@ -12,8 +12,6 @@ def test_code_blocks_wrap_instead_of_overflowing_the_column():
     assert "max-width: 100%" in css
 
 
-
-
 def test_svg_is_constrained_to_the_column_width():
     """An inline <svg> pulled in from article HTML must not render at its own unconstrained
     width/height - see the corner-bracket icon that rendered most of a page tall/wide in a
--- a/goosepaper/test_styles.py
+++ b/goosepaper/test_styles.py
@@ -1,15 +1,6 @@
-from .styles import Style
 import re
 
-
-def test_code_blocks_wrap_instead_of_overflowing_the_column():
-    """A <pre> block wider than its column must wrap, not overflow into whatever renders next to
-    it - see the multi-column overlap this was fixed for."""
-    css = Style("FifthAvenue").get_css(page_profile="paper_pro", layout="2col")
-
-    assert "white-space: pre-wrap" in css
-    assert "overflow-wrap: break-word" in css
-    assert "max-width: 100%" in css
+from .styles import Style
 
 
 def test_svg_is_constrained_to_the_column_width():
--- a/goosepaper/styles.py
+++ b/goosepaper/styles.py
@@ -282,6 +282,36 @@ def _base_print_css(
         height: auto;
     }}
 
+    /* `<pre>`'s default `white-space: pre` never wraps long lines - fine in a browser (you get a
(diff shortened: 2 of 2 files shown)


goosepaper/test_styles.py
```diff
@@ -1,6 +1,17 @@
+from .styles import Style
 import re
 
-from .styles import Style
+
+def test_code_blocks_wrap_instead_of_overflowing_the_column():
+    """A <pre> block wider than its column must wrap, not overflow into whatever renders next to
```
j6k4m8: I don't totally understand the reason to test for specific strings; is there an expectation of drift?

Smengerl: Fair point. To be upfront: this pattern isn't an established convention from you or other contributors — the other tests in this file that do the same thing (test_svg_is_constrained_to_the_column_width, test_unsized_svg_defaults_to_icon_size, test_appendix_block_gets_exactly_one_page_break_not_per_entry) are mine too. So no, there's no expectation that these exact strings will drift — I'm using string assertions across several of my PRs specifically because they're cheap to write and catch the case I actually care about: someone (including future-me) accidentally deleting or breaking the CSS rule during a refactor, without needing to spin up a full PDF-rendering check. It's a regression guard for "is the fix still wired in," not a proof that the layout renders correctly — that part I verified manually via the screenshots in the PR description.

Happy to make it less brittle against pure formatting changes — e.g. re.search on a looser pattern instead of an exact substring, like the appendix test already does — if that addresses the concern better than a straight in css check.

j6k4m8: Let's just drop those tests, it'll be obvious if we drop the css in a diff :)

j6k4m8: I think with this change we can merge this pr asap :)

j6k4m8 (changes requested):

Smengerl:
Dropped the brittle string-assertion test — the CSS diff will make it obvious if the fix gets reverted, as you said. Pushed in 7e8e168.

Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #126 — both insert a new CSS block at the same anchor point in `styles.py`'s `_base_print_css`. Suggested order: this one before #126.

PR #121, opened 2026-07-29
Add native content_filters/skip_title_patterns to "rss" sources

Smengerl:
## Why

goosepaper fetches an RSS entry's summary, but for real article content it follows the entry's link and runs `readability` against the full page - and that's where the trouble starts. `readability` extracts "the article", not
"the article as the author wrote it": it keeps whatever DOM structure the page's CMS wrapped the actual text in, and most news sites wrap it in a lot.
Cookie/consent banners, gallery/lightbox widgets, "read more"/"show less" teaser toggles, newsletter signup CTAs, related-content boxes, ad slot placeholders - none of that is prose, all of it survives extraction, and all of it printed straight onto the newspaper page with no way to remove it before this PR.

This isn't hypothetical - I built this feature to run my own daily newspaper (~25 RSS feeds) and needed it from day one to keep the output readable. Some concrete examples from real feeds, all currently running in production:

| Feed | Problem | Filter used |
|---|---|---|
| heise online / heise iX / heise Developer | Image gallery widgets, mobile ad wrapper divs, a cookie-notice `<details>`, gift-article/opt-in custom elements, "Mehr/Weniger anzeigen" toggle text | `{"type": "css", "selector": "div.Gallery"}`, `div[class*='ad-mobile-group']`, `details.notice-banner`, `a-gift`, `a-opt-in`, plus `{"type": "regex", "pattern": "Mehr anzeigen"}` / `"Weniger anzeigen"` |
| heise (all three feeds above) | Sponsored posts, paywalled heise+ articles, vendor deal posts - not something you can strip out of one article's HTML, the *whole entry* needs to go | `skip_title_patterns: ["^anzeige:", "^heise\\+ \\|", "^heise-angebot:"]` |
| t3n | Cookie-consent widget, a "suggested articles" box, a newsletter-signup CTA button | `div[data-consent-service-id]`, `div.c-suggestNews-container`, `a.t-newsletter-singleview-top-cta` |
| Electrek | Outbrain ad label + ad slot placeholder, plus a boilerplate "considering going solar / FTC disclosure" paragraph at the end of every single article | two `css` selectors, plus `{"type": "regex", "pattern": "(considering going solar\|FTC: We use income earning).*", "flags": "s"}` (the `s` flag so `.` also matches the newlines in that trailing block) |
| CNBC | A "related content" widget and an ad-slot `div` identified only by a `data-module` attribute | `div.RelatedContent-relatedContent`, `[data-module='mps-slot']` |
| netzpolitik.org | A donation/support call-to-action box repeated in every article | `div.netzpolitik-cta` |
| mobiFlip | A "read more" pill-shaped link at the end of every teaser | `a.gf-pill` |
| Hacker News (via hnrss) | Anchor-jump icons (SVG) inside comment-count links - not content, just chrome | `a[href^='#']:has(svg)` (also proves the CSS `:has()` selector works through this pipeline's selector engine - `soupsieve`) |
| GameStar | A consent-management-platform wrapper div around the actual text | `div#cmp-content` |
| Car and Driver | Whole articles that are just a photo gallery caption, not readable prose in a print layout | `skip_title_patterns: ["^view photos of"]` |

Doing this as a wrapper *around* goosepaper isn't really possible: the
cleanup has to run on the exact HTML `RSSFeedStoryProvider` already
extracted via `readability`, inside `get_stories()` - an external layer has
no hook into that step without re-implementing the whole fetch/extract
pipeline itself. So it belongs here, as new optional, plain-JSON fields
on the `"rss"` source type, no custom code required to use them.

**Grown since first opened:** this PR started out with just a `content_filters` denylist and `skip_title_patterns`. Review/production use surfaced the need for the allowlist direction too (keep only what matches, not just drop what doesn't), so `content_filters` was renamed to `skip_content_filters` once its allowlist counterpart `accept_content_filters` existed - "skip" alongside a bare "content_filters" read as if it were the only direction. `min_body_text_length`/`max_body_text_length` were added for the same production feeds, to catch extraction failures and outlier-length articles that no selector/regex rule targets. The table above and the rest of this description reflect the current, full scope.

## What

RSS sources can filter along two independent axes - what to match (title or fetched article content) and which direction (skip = denylist, accept = allowlist):

|         | Skip (denylist)          | Accept (allowlist)         |
|---------|---------------------------|-------------------------------|
| Title   | `skip_title_patterns`    | `accept_title_patterns`     |
| Content | `skip_content_filters`   | `accept_content_filters`    |

- **`skip_title_patterns` / `accept_title_patterns`** - flat list of regexes, matched case-insensitively against the entry title, before it's even fetched. `skip`: a match drops the entry (e.g. `["^anzeige:", "^sponsored"]` to drop sponsored posts). `accept`: if non-empty, only matches are kept (e.g. `["amazon", "amzn"]` to build a single-company news ticker out of an otherwise general feed).
- **`skip_content_filters`** - list of `{"type": "css", "selector": "..."}` (deletes matching elements, e.g. ad blocks or cookie banners) or `{"type": "regex", "pattern": "...", "flags": "i"}` (strips matching text; `flags` optional, any of `i`/`s`/`m`/`x`) rules. `type` decides which other keys are valid - a `css` entry can't carry `pattern`/`flags`, a `regex` entry can't carry `selector`. Regex rules run first, then CSS rules, regardless of list order.
- **`accept_content_filters`** - list of `{"type": "css", "selector": "..."}` or `{"type": "regex", "pattern": "...", "flags": "i"}` rules, tried/applied independently of each other. `css`: the first selector that matches wins and the article is replaced with just that element's contents - useful when `readability`'s own extraction misses and you know exactly which container holds the real content; no match leaves the article unchanged. `regex`: a whole-story gate rather than a transform - matched against the fetched article's text (not raw markup), so a story is kept only if it matches at least one `regex` filter; a `css` filter can't sensibly gate this way, since a miss should leave the article as-is rather than drop it.
- **`min_body_text_length` / `max_body_text_length`** - optional, applied after the content filters above; drop stories whose extracted body's visible text length falls outside that range. `min` catches a failed extraction (a near-empty body); `max` catches an implausibly long body (e.g. a hardware review with a huge photo gallery/spec dump) that would otherwise balloon a single entry into the bulk of the whole paper.

## Reproduction

A feed returning two entries - one with a sponsored-post title, both
sharing the same messy article HTML (an ad block, a cookie banner, and a
trailing "Mehr anzeigen" link, modeled directly on the heise example above):

**Before (master):** no way to express this in config at all -
none of the fields above are recognized parameters. Both
entries come through untouched, junk included:

2 stories returned:

--- Anzeige: Sponsored post you'd rather skip ---
<p>Real article content starts here.</p><div class="ad-block">BUY NOW! Limited offer!</div><div class="cookie-banner">We use cookies. Accept all?</div><p>More real content in the middle.</p><p>Mehr anzeigen</p>

--- A real headline about something interesting ---
<p>Real article content starts here.</p><div class="ad-block">BUY NOW! Limited offer!</div><div class="cookie-banner">We use cookies. Accept all?</div><p>More real content in the middle.</p><p>Mehr anzeigen</p>

**After (this branch)**, with:
```json
"skip_content_filters": [
  {"type": "css", "selector": "div.ad-block"},
  {"type": "css", "selector": "div.cookie-banner"},
  {"type": "regex", "pattern": "Mehr anzeigen"}
],
"skip_title_patterns": ["^anzeige:"]
```

1 story returned:
--- A real headline about something interesting ---
<html><body><p>Real article content starts here.</p><p>More real content in the middle.</p><p></p></body></html>

The sponsored entry is gone entirely, and the ad block/cookie banner/tracking text are stripped from the survivor. (The <html><body> wrapper and the leftover empty <p></p> are BeautifulSoup's own re-serialization once at least one CSS filter runs - cosmetically harmless, WeasyPrint renders it the same either way.)

## Implementation
* `goosepaper/contentfilters.py` (new): `apply_skip_content_filters()` / `apply_accept_content_filters()` / `should_skip_title()` / `should_accept_title()` / `should_accept_content()` / `visible_text_length()` - the actual filtering logic, framework-independent.
* `goosepaper/storyprovider/rss.py`: `RSSFeedStoryProvider` applies title-level filters first thing in its `get_stories()` loop (before any fetch), then content-level filters and the body-length gate on the extracted `body_html` right after fetching - regardless of which `body_source` mode produced that HTML.
* `goosepaper/config.py`, `goosepaper/util.py`: schema validation and provider wiring for all six new optional `"rss"` fields.

## Testing
* `test_contentfilters.py` (new): unit tests for every function above (css/regex removal and gating, flags, ordering, body-length thresholds).
* `test_rss.py`: integration tests proving `RSSFeedStoryProvider` actually applies each new field to a fetched story, and that title-level skip/accept happens before an entry counts toward `limit`.
* `test_config.py`: schema validation for all six fields, including the error message for each malformed case (unknown filter type, missing selector/pattern, cross-type fields, unknown fields).
* Full test suite passes (120 passed as of this branch's current commit).
* `example-config.json` (updated with working examples of every field) generates a real PDF end-to-end.
* Every filter in the table above is running against its real feed in production, in a daily-generated newspaper, right now.

## Scope
Additive only - all six fields are optional, existing "rss" sources without them are completely unaffected.

## Merge overlap note

Verified by locally merging every pairwise combination of my currently open PRs against `master`. This one produces a real but purely mechanical (non-semantic) merge conflict with **#123, #124, #127, #128, #129, #132, #133** - all add entries to one of the same shared registries this PR also touches (`config.py`'s per-field validator dict, `util.py`'s provider dispatch dict, or `rss.py`'s `RSSFeedStoryProvider.__init__`/`get_stories`). In every case checked, the conflicting insertions are independent (different field/provider names, no logic overlap) - resolving means keeping both sides, not choosing between them.


goosepaper/storyprovider/rss.py
```diff
@@ -22,6 +23,8 @@ def __init__(
         since_days_ago: int = None,
         byline: str = "all",
         body_source: str = "auto",
+        content_filters: Optional[List[Dict[str, Any]]] = None,
```
j6k4m8: Nomenclature question for you — what do you think of something like "`content_skip_filters`" to distinguish "filter-in" vs "filter-out"? For example, "`skip_title_patterns`" feels very obvious (these are the titles to skip) but "content filter" feels like maybe I _do_ want things that match, but I think the code here filters the matchers _out_. 

Thoughts? Otherwise LOVE the idea!

j6k4m8: @Smengerl this also makes me wonder if we want to support both `filter_reject` and `filter_accept` in case you know that the actual story is inside a particular selector and the junk is unpredictable...

Smengerl: I can see title_filter_accept being useful (e.g. a stock ticker looking for one company's name). 

For content, I'd distinguish the two filter types: 
* a regex accept doesn't make much sense to me — the result would essentially be whatever static phrase the pattern matches, not coherent prose. 
* A CSS-selector accept, though, is a real pattern (select_one the container and keep only that instead of the whole parsed tree) — useful exactly for the case you describe, where readability's own extraction misses and you know precisely which container holds the real article. I didn't need it for any of my own feeds, but happy to add it if you think it's worth including now vs. as a follow-up.

Smengerl: I'll rename content_filters to content_skip_filters as you suggested, and leave skip_title_patterns as-is since you're happy with it. For future symmetry when we add accept-mode, I'd go with accept_title_patterns and content_accept_filters (matching each existing field's own word order) rather than forcing both families into the same shape. I'll push the rename to this branch shortly.

Smengerl: Pushed the rename as promised, plus the accept-mode fields we discussed:

* content_accept_filters / accept_title_patterns are implemented, using the naming scheme from my last comment. accept_title_patterns is the allowlist counterpart to skip_title_patterns (same shape, just kept instead of dropped). content_accept_filters is CSS-only — no type field, since a regex "accept" would just reduce a story to whatever static phrase the pattern matches, not real prose. It tries each selector in list order, keeps the first match's contents, and falls through unchanged if nothing matches (a miss shouldn't zero out an article).
* Tested it against the the stock ticker scenario: accept_title_patterns: ["amazon", "amzn"] against Google News' general Business feed (70 entries) correctly narrows to the 4 that are actually about Amazon.
* Found and fixed a validation gap in my previous commits: content_skip_filters used one combined allowed-keys set for both "css" and "regex", so e.g. {"type": "css", "selector": "...", "pattern": "unused"} passed validation silently instead of erroring. It's per-type now.
* Also renamed the internal apply_content_filters() helper to apply_content_skip_filters(). It kept its pre-rename name when the config field became content_skip_filters (deliberately, to keep that change small), but with apply_content_accept_filters() now sitting right next to it, the omission read as accidental rather than intentional. Internal-only, no config-facing change.
* Restructured the README section about the filter usage — it had grown into three dense paragraphs. It's now a skip/accept × title/content table plus one bullet per field.

j6k4m8: sorry to make you do an extra round trip but I think it's probably worth keeping the word order consistent,

`skip_content_filter` and `skip_title_filter`, OR `content_skip_filter` and `title_skip_filter`, doesn't make sense to have them flips of each other, unless I'm not following your logic...

Regex for accept I think would be "anything with this regex passes"; e.g., like you mention, any story matching `/AAPL/` >0 times...  Thoughts on that? I think this is an awesome feature!

Smengerl: Good point on both counts, thanks — pushed a rename + the regex accept support.

**Naming**
Went with skip_content_filters / accept_content_filters, so all four fields now share the same verb-first order: skip_title_patterns, accept_title_patterns, skip_content_filters, accept_content_filters.

**Regex for accept**
Added a "regex" type to accept_content_filters. It works differently from "css" on purpose: css narrows the kept content down to one container, but a regex can't sensibly do that (matching a fragment isn't "the container"), so a regex entry is a whole-story gate instead — matched against the fetched article's extracted text (not raw markup, so it can't false-hit inside a tag/attribute). If nothing matches, the entry is dropped entirely, same as accept_title_patterns but at the content level:

"accept_content_filters": [
  { "type": "regex", "pattern": "AAPL" }
]
Tested against a general business feed - keeps only entries whose body actually mentions the ticker, title notwithstanding.

One side effect of adding "type" to accept_content_filters: existing {"selector": "..."} entries without a type field now need "type": "css" added explicitly, since type is required to disambiguate from regex. Called out since it's a breaking format change on this field (nothing's shipped yet, so should be a no-op for anyone tracking master).

Smengerl:
Heads up: I just pushed an additional commit to this PR that's a bit of scope creep beyond the original title — apologies in advance.

Alongside the content/title filters, I added two more optional `"rss"` source fields: `min_body_text_length` and `max_body_text_length`, which drop stories whose extracted body's visible text length falls outside a given range (catching failed extractions on the low end, and outlier articles — e.g. I recently had a gigantic hardware review in heise RSS — ballooning the output by more than 200 pages). Also extremely short stories in a feed usually indicate that there is no reasonable content to expect or its another form of add that was not filtered out by the title/content filters.

I bundled it into this PR rather than opening a separate one because it depends on the same `contentfilters.py` module this PR introduces (`visible_text_length` lives there, applied at the same point as the other filters, right after `skip_content_filters`/`accept_content_filters`). Splitting it out would have meant introducing that module twice.

Happy to pull it back out into its own PR on top of this one if you'd rather review/merge them separately — just let me know.

Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR touches both `storyprovider/rss.py` and `config.py`'s source-schema block, so it conflicts with the most siblings of anything in the queue: #129, #128, #127 (rss.py) and #133, #132, #124 (config.py) — 6 total. Recommend merging this one last overall, after both the rss.py group and the config.py group land, so it only needs one final rebase instead of repeated ones.

Smengerl:
Heads-up, separate from the sibling-conflict note above: GitHub currently reports this PR as `dirty` against `master` — caused by #138 (`register_story_provider`, merged 2026-08-08, not one of my PRs) touching `test_config.py` at the same spot this PR's own additions land. Trivial rebase, unrelated to the sibling-PR conflict clusters.

PR #122, opened 2026-07-29
Add PlacementPreference.APPENDIX for reliable end-of-document content

Smengerl:
## Why

There's no way for a story provider to say "this content belongs at the
very end of the document, in its own clearly separated section" -
`PlacementPreference.FULLPAGE` looks like the answer (`break-before: page`
on the story), but that CSS property doesn't reliably start a new page for
a fragment rendered *inside* a multi-column container in this WeasyPrint
version. In practice that means FULLPAGE content just continues wherever
there happens to be room in the current column - squeezed into a
2-/3-column layout, indistinguishable from a regular article, with no
guaranteed separation from whatever precedes it.

I ran into this building a `"puzzle"` story provider (a separate PR):
solutions need to land reliably at the very end of the issue, grouped
together like a print puzzle book's answer section - not wherever
FULLPAGE happens to land them in a multi-column layout.

Note: this is **provider-level API**, not something exposed as a JSON
config option - it's a `Story.placement_preference` value any story
provider's Python code can set, the same way `EAR`/`SIDEBAR`/`UTILITY`
already are.

Before PR:
image-1785359546670

With PR:
image-1785360203420
image-1785360207931

## What

Adds `PlacementPreference.APPENDIX` (`goosepaper/util.py`), a fourth
story region alongside the existing EAR/SIDEBAR/UTILITY mechanism.
Stories marked with it are pulled out of normal narrative order and
rendered together in their own `.appendix` block, positioned *outside*
the `column-count` container entirely - after `.main-stories` and
`.sidebar`, at the very end of the document. Because it never enters the
multi-column layout in the first place, it's reliably visually distinct
regardless of column count - not dependent on a CSS break property that
doesn't work reliably in that context.

- `goosepaper/util.py`: new `PlacementPreference.APPENDIX` enum member.
- `goosepaper/story.py`: maps it to the `"placement-appendix"` CSS class.
- `goosepaper/styles.py`:
  - `.appendix { break-before: page; }` - one guaranteed page break before
    the block as a whole starts, so it never begins mid-page tacked onto
    whatever short content happened to precede it.
  - `.appendix > article` gets the same divider/spacing styling as
    `.main-stories`/`.sidebar` articles, flowing one after another rather
    than forcing a page break *per story* - deliberately not the same
    rule as above: a break on every article would waste a page per entry
    for something like a single puzzle solution (an earlier version of
    this branch did that, then walked it back - see the second commit).
- `goosepaper/goosepaper.py`: partitions APPENDIX stories out of
  `main_story_objects`, renders them via the existing
  `_render_story_region` helper (so anchor IDs, TOC entries, and story
  numbering all work exactly like any other region), and appends the
  result after the main stories/sidebar block, outside the columns
  container.

## Reproduction

**Column breakout** - a 2-column paper with three filler articles plus one
extra story meant to land in its own end section, compared using the
closest existing equivalent, `FULLPAGE`, on `master`, versus `APPENDIX` on
this branch:

- **Before (master, `FULLPAGE`):** the extra story renders squeezed into
  the second column, right after the third article - no separation, no
  distinct section, just more column-flowed content.
- **After (this branch, `APPENDIX`):** the same content breaks out to full
  page width, spanning both columns - immediately readable as a distinct
  end-of-document section rather than another squeezed-in column article.

**Single guaranteed break, not per entry** - a short main article (with
plenty of room left on its page) followed by two short appendix entries renders as exactly 2 pages - page 1 has just the main article, page 2 has *both* appendix entries together, sharing the page rather than each
  forcing its own break. Confirms the block-level break fires once, and
  the previous "one page per entry" behavior (which this branch's second
  commit deliberately removed) doesn't come back.

Screenshots of the column-breakout comparison attached to this PR.

## Testing

- `test_goosepaper.py`:
  - `test_appendix_stories_render_after_stories_in_their_own_block` -
    asserts the `.appendix` block renders after `.stories` in the HTML and
    that appendix stories aren't duplicated into `.main-stories`.
  - `test_appendix_stories_with_identical_headline_are_deduplicated` -
    proves `Goosepaper.get_stories(deduplicate=True)` (no new dedup
    mechanism needed) correctly collapses two same-headline APPENDIX
    stories from different providers down to one, the way a shared
    "how does Sudoku work?" endnote requested by several puzzle sources
    should.
- `test_styles.py`:
  - `test_appendix_block_gets_exactly_one_page_break_not_per_entry` -
    asserts `break-before: page` exists on `.appendix` and specifically
    does not exist on `.appendix > article`.
- Full test suite passes (78 passed - includes the already-merged
  Wikipedia/SVG-overflow fixes, since this branch merges cleanly with the
  current `master` tip; verified directly, no conflicts despite both
  touching `styles.py`).
- Both reproductions above were run directly against real PDF output
  (page counts and text-position checks via PyMuPDF), not just asserted
  in unit tests.

## Scope

Pure addition to the existing placement-preference mechanism - no
existing `PlacementPreference` value's behavior changes, so this can't
affect any current story provider. A puzzle story provider PR that
consumes this for solution placement will follow separately.

### DIFF
--- a/goosepaper/goosepaper.py
+++ b/goosepaper/goosepaper.py
@@ -121,6 +121,7 @@ class Goosepaper:
                 PlacementPreference.EAR,
                 PlacementPreference.SIDEBAR,
                 PlacementPreference.UTILITY,
+                PlacementPreference.APPENDIX,
             ]
         ]
 
@@ -135,8 +136,17 @@ class Goosepaper:
             for story in stories
             if story.placement_preference == PlacementPreference.SIDEBAR
         ]
+
+        appendix_story_objects = [
+            story
+            for story in stories
+            if story.placement_preference == PlacementPreference.APPENDIX
+        ]
         ordered_story_objects = (
-            utility_story_objects + main_story_objects + sidebar_story_objects
+            utility_story_objects
+            + main_story_objects
+            + sidebar_story_objects
+            + appendix_story_objects
         )
         story_anchor_ids = self._story_anchor_ids(ordered_story_objects)
         story_numbers = {
@@ -162,8 +172,14 @@ class Goosepaper:
             story_numbers,
             used_anchors=used_anchors,
         )
+        appendix_stories, appendix_toc_entries = self._render_story_region(
+            appendix_story_objects,
+            story_anchor_ids,
+            story_numbers,
+            used_anchors=used_anchors,
+        )
         toc_html = self._render_table_of_contents(
-            utility_toc_entries + main_toc_entries + sidebar_toc_entries,
+            utility_toc_entries + main_toc_entries + sidebar_toc_entries + appendix_toc_entries,
             enabled=table_of_contents,
             effective_columns=effective_columns,
         )
@@ -191,6 +207,13 @@ class Goosepaper:
             utility_html = f"""
                     <div class="utility-strip">
                         {''.join(utility_stories)}
+                    </div>
+            """
+        appendix_html = ""
+        if appendix_stories:
+            appendix_html = f"""
+                    <div class="appendix">
+                        {''.join(appendix_stories)}
                     </div>
             """
 
@@ -246,6 +269,7 @@ class Goosepaper:
                     </div>
                     {sidebar_html}
                 </div>
+                {appendix_html}
             </body>
             </html>
         """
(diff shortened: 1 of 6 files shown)


j6k4m8:
Another banger! This is a great idea. Also it's good to know that the FULLPAGE option doesn't quite work all the time as intended... Merging this one now! Thank you @Smengerl, I'm loving this PR day!

PR #123, opened 2026-07-29
Add a "puzzle" story source (Sudoku, Binoxxo, Futoshiki, Kakuro, Shikaku)

Smengerl:
## Why

I read Goosepaper on a reMarkable every morning, and wanted an actual
newspaper-style logic puzzle page in there - the kind of thing print
newspapers have always had, that a pure news feed doesn't. There was no
way to express "give me a Sudoku" in a goosepaper config at all, so this
PR adds a full puzzle generator + renderer as a new source type, native to
goosepaper rather than bolted on externally (a wrapper has no way to add
a genuinely new *kind* of content - only new config for existing story
providers).

This isn't a proof of concept - it's running in production across three
real, daily-generated newspapers I maintain, generating and delivering a
mix of all five puzzle types at every difficulty every day.

## What

- **Generation**: self-contained grid/solver code for five puzzle types
  (Sudoku, Binoxxo, Futoshiki, Kakuro, Shikaku) under `goosepaper/
  puzzlegen/` - no `reportlab`, no third-party puzzle-generation package,
  no rendering or CLI code, just the pure generate/verify-unique-solution
  logic per type.
- **Rendering**: plain HTML `<table>`s with inline CSS (`goosepaper/
  storyprovider/puzzle.py`) - no image rasterization, renders natively
  through WeasyPrint like everything else in goosepaper.
- **Config surface**: a `"puzzle"` source with:
  - `puzzle_type` (required) - one of `sudoku`, `binoxxo`, `futoshiki`,
    `kakuro`, `shikaku`. No default on purpose - a config that forgets it
    should fail loudly instead of silently always generating Sudoku.
  - `box_size` (Sudoku only, default `3`) - `2` for a quick 4x4 variant or
    `3` for the classic 9x9. Doesn't vary by difficulty.
  - `difficulty` (default `"medium"`) - `easy`/`medium`/`hard`. For every
    type but Sudoku, this also determines the grid size (e.g. Shikaku's
    "hard" preset is specifically tuned for a 20x20 grid), so grid size
    and difficulty are deliberately not two independent settings a config
    could mismatch.
  - `count` (default `1`) - how many instances of this type+difficulty.
  - `seed` (optional) - for reproducible generation.
  - `name` (optional) - a visible per-instance heading; omit it to rely on
    the enclosing section's own heading instead.

## Reproduction

**Basic generation**, one source with `count: 2`:

```json
{
  "type": "puzzle", "puzzle_type": "sudoku", "difficulty": "medium",
  "count": 2, "name": "Sudoku"
}
```

*Before (master):*
Honk! Unknown source type "puzzle". 

*After (this branch):*
2-page PDF - page 1 has both Sudoku grids side by side, page 2
has both solutions together, batched at the end rather than interleaved
with their own puzzle. Screenshots attached.

image-1785361409032
image-1785361413858
image-1785361417995

## Known limitation (documented in code, worth flagging here too)
Solution batching only reliably groups solutions within one provider
instance. With two separate "puzzle" sources (e.g. one Sudoku, one
Binoxxo - the realistic multi-type case), each provider's own solution
still lands right after that provider's own puzzle, not collected
together at the very end of the whole issue - screenshot attached showing
exactly this with two sources. 
The root cause: PlacementPreference. FULLPAGE's break-before: page doesn't reliably force a new page inside
a multi-column layout in this WeasyPrint version (verified in isolation),
so "batch at the end" only works by accident, within what one provider
itself returns in order.

I already have the real fix for this as a separate, smaller, fully
independent PR: PlacementPreference.APPENDIX (#122), which collects
appendix-tagged content from every provider into one shared block at
the true end of the document. A follow-up PR switches this puzzle
provider's solutions over to that mechanism once #122 lands - I split it
out this way so each piece can be reviewed on its own rather than as one
large combined diff.

## Testing
* 88 tests pass (puzzle generation/rendering for all 5 types, config-
schema validation and provider wiring for puzzle_type/box_size/
difficulty/count/seed/name, headline-disambiguation for
count > 1, and the size-derived-from-difficulty behavior).
* Verified end-to-end against three real newspaper configs I run daily
(real PDF generation, not just unit tests), including a real delivery
run to a reMarkable.
* Both reproductions above were run fresh against master and this
branch to produce the results shown.

## Scope
Additive only - a brand new source type, no existing behavior changes.

## Merge overlap note

Verified by locally merging every pairwise combination of my currently open PRs against `master`. This one produces a real but purely mechanical (non-semantic) merge conflict with **#121, #132, #133** - all add a new source type/fields to the same shared registries in `config.py`/`util.py`. (#124 is the one exception - see its own PR description for why: it's a dependency, not a sibling conflict, and merges cleanly on top of this one.)


Smengerl:
## Changes since the last merge

**Shikaku** — cells are now as large as Sudoku's (`2em`/`1.15em`) wherever that fits, keyed off both difficulty (5x5/10x10/20x20) and column count, since a single flat size can't fit all three:

| Difficulty | 1-col | 2-col | 3-col |
|---|---|---|---|
| Easy | 1.15em | 1.15em | 1.15em |
| Medium | 1.0em | 0.9em | 0.6em |
| Hard | 0.55em | 0.4em | 0.3em |

*[Screenshots: easy/medium/hard x 1/2/3-col]*
image-1785760190936
image-1785760171311
image-1785760196499
image-1785760176166
image-1785760202354
image-1785760207038
image-1785760212372
image-1785760216887
image-1785760221499


Sudoku-size everywhere doesn't work for the larger grids: a 20x20 grid overflows every column width and is even tall enough to break across pages in a single column. Each tier is also picked so the table's own natural width fits the column without being squeezed - otherwise cell height (fixed) and width (auto-shrunk to the column) diverge and cells render as rectangles instead of squares, as briefly happened with Medium in 3-column before these tiers existed.

**Kakuro** — two small contrast fixes to clue cells: the diagonal fill was black-with-a-gray-line-and-white-text (inverted from the standard gray-fill/black-line/black-text look, hard to read), and clue cells' black outline was getting visually lost against neighboring cells' lighter gray border. Both now consistent.

**Futoshiki** — empty inequality-gap cells (no `<`/`>` relation to show) had no content at all, so their row collapsed shorter than rows that did show a relation. Filled with `&nbsp;` and pinned line-height so every gap row measures the same height.

All verified via actual WeasyPrint renders (page count, and per-cell width/height measured from the rendered PDF), not just computed from the CSS values.

j6k4m8:
This is a great addition and I have a feeling we'll have a lot of people who are very excited to have this!

It looks to me like you had LLM help writing this PR (totally cool with me) — can you confirm that this code does not overlap with any existing implementations / can be released under the Goosepaper Apache 2.0 license? I know that for a few of these puzzle types (at least sudoku) there are widely-regarded algorithms but these have varying levels of licensing/protection.

Also — I am not an expert on any of these puzzle types; can we verify that these are all uniquely and fully solvable from the provided clues? That's my only other "gotcha" before merging. 

If either of these are a "maybe" my preference is to (at least temporarily) put these into [Auxiliary Goose](https://github.com/j6k4m8/auxiliary-goose/) where they can be tested and used by users ([the official goosepaper app](https://goosepaper.jordan.matelsky.com/), as well as a few others, inherit both from this repo as well as AG) but we don't make guarantees of stability/back-compatibility.

SUPER cool work, excited to add these to my daily, wherever they wind up :)

j6k4m8:
PS @Smengerl thank you a ton for this great feature push, this is a huge improvement to the Goosepaper codebase and I am super appreciative! Thank you for your patience with my reviews :)

Smengerl:
# Unique Solvability of the Goosepaper Logic Puzzles

I've been using this feature for a while now and have always gotten uniquely solvable puzzles — here's the reasoning behind that, straight from the code.

## Shared principle
Every generator first builds a complete, valid solution and only ever modifies it under one condition: a change is **kept only if a full backtracking counter (stopping as soon as a 2nd solution is found) still reports exactly 1 solution afterward.** If that check fails, the change is reverted or a fresh attempt is made. This isn't a heuristic — it's an exhaustive search.

## Sudoku
`generate_puzzle` removes numbers one at a time, in random order, from the full solution. After each removal, `has_unique_solution` (backtracking + naked-/hidden-single propagation) checks the remaining grid; if it's still unique, the cell stays empty, otherwise the value is written back.
See `goosepaper/puzzlegen/sudoku/puzzle.py`, `state.py`

## Futoshiki
Same mechanism as Sudoku, with the fixed inequality constraints additionally taken into account by the solver. The `<`/`>` relations are first derived from the solution and locked in, then the same remove-and-check loop runs.
See `goosepaper/puzzlegen/futoshiki/puzzle.py`, `solver.py`

## Binoxxo
Also remove-and-check from a full solution. One wrinkle: the search has a node budget for large boards (14×14); if that budget is exceeded, the result is **conservatively treated as "not unique"** rather than guessed — when in doubt, the removal is discarded.
See `goosepaper/puzzlegen/binoxxo/puzzle.py`, `state.py`

## Shikaku
No clues are ever removed — every rectangle contributes exactly one clue by definition. Instead, after generating the rectangle partition, the *entire* clue set is checked: if the same numbers could be tiled a different way (or the check exceeds its search budget), the whole partition is discarded and a fresh one is drawn (up to 300 attempts).
See `goosepaper/puzzlegen/shikaku/puzzle.py`, `solver.py`

## Kakuro
The reverse approach: classic Kakuro has no numeric givens, only sum clues. It's first checked with no givens at all; if that's ambiguous, individual solution cells are revealed one by one until uniqueness is reached — capped at 30% of cells revealed, beyond which the layout is discarded rather than forcing a "broken" puzzle.
See `goosepaper/puzzlegen/kakuro/puzzle.py`, `solver.py`

Smengerl:
I noticed EbookLib (AGPLv3+) and pyphen/crc32c (LGPL) are already transitive dependencies of the existing project (via ebooklib/weasyprint) — that predates this PR and isn't something the puzzle-generation code introduces or depends on. The puzzlegen modules themselves add zero new dependencies.

Smengerl:
Re: license/originality — I ran two checks to double-check this:

ScanCode Toolkit (license + copyright fingerprint scan) against the full puzzlegen/ source tree (76 files): 0 files with any license or copyright match: 
[license_scan.sh](https://github.com/user-attachments/files/30672853/license_scan.sh)

GitHub code search for several distinctive, verbatim phrases from the docstrings/comments in the puzzle generators (e.g. exact wording from the Sudoku/Kakuro/Binoxxo module docstrings) — no matches anywhere on GitHub. Generic names like SudokuState do show up in many unrelated repos, but that's just a common, obvious class name for the concept, not code overlap.
No new dependencies were introduced by the puzzle-generation code either — it's pure standard-library Python. The techniques used (backtracking + constraint propagation, naked/hidden singles, etc.) are standard, decades-old CS textbook algorithms, not tied to any specific implementation.

Happy to share the raw scan output if useful.


---

Scan Code result:
```
Bootstrapping ScanCode (first run only takes a while)...
Scanning goosepaper/puzzlegen ...
Setup plugins...
Collect file inventory...
Scan files for: info, licenses, copyrights with 4 process(es)...
[####################] 128                                  
Scanning done.
Summary:        info, licenses, copyrights with 4 process(es)
Errors count:   0
Scan Speed:     5.35 files/sec. 19.90 KB/sec.
Initial counts: 76 resource(s): 64 file(s) and 12 directorie(s) 
Final counts:   76 resource(s): 64 file(s) and 12 directorie(s) for 237.96 KB
Timings:
  scan_start: 2026-08-03T174738.470121
  scan_end:   2026-08-03T174753.128819
  setup_scan:licenses: 2.58s
  setup: 2.58s
  inventory: 0.11s
  scan: 11.96s
  total: 14.68s
Removing temporary files...done.

Done. Summary of any hits:

0 of 76 scanned files have a license/copyright hit.
No known third-party license text or copyright notices detected.
```

Smengerl:
Closing per your request on #124 — #124's branch already fully contains this PR's code (it was branched off this one and kept in sync via merge ever since; `goosepaper/puzzlegen/` is byte-identical between the two). #124 is now the single PR for all puzzle logic (provider + solutions/explanations). Merge-order context for the rest of the open PR queue is on #124 as well.

PR #124, opened 2026-07-29
Add puzzle story source (Sudoku, Binoxxo, Futoshiki, Kakuro, Shikaku) with solutions in the appendix and optional explanations

Smengerl:
## This PR now covers the whole puzzle feature (combined with #123)

Per request: this PR is now the single place for all puzzle logic — the `"puzzle"` source itself (Sudoku, Binoxxo, Futoshiki, Kakuro, Shikaku, from #123) plus the appendix/explanations work this PR originally added on top of it.

Nothing needed to change code-wise to get here: this branch (`feature/puzzle-explanations`) was branched off #123's branch and has been kept in sync with it via merge ever since, so it already fully contains #123 — `git diff` between the two branches on `goosepaper/puzzlegen/` is empty. I've closed #123 as superseded by this PR.

Both other former dependencies are also resolved: #122 (`PlacementPreference.APPENDIX`) and #130 (`deduplicate` wired through `to_html()`/`to_pdf()`) are merged, so the "Correction" caveat further down about deduplication not reaching the normal CLI/Docker path no longer applies — it does now.

## Why

Two things this PR fixes, both problems with the puzzle provider as it stood on its own before this PR:

**1. Solutions don't actually collect at the end with multiple puzzle
sources.** Batching via `PlacementPreference.FULLPAGE` alone only reorders *within one
provider's own output*. With two separate `"puzzle"` sources (the
realistic case - a real newspaper mixing Sudoku, Binoxxo, etc.), each
provider's solution would land right after that provider's own puzzle,
not collected together at the true end of the issue.

Reproduced directly: two-source config (one Sudoku source, one Binoxxo source) - before,
each solution appeared immediately next to its own puzzle; after, both
puzzles render together first, then both solutions together in the
appendix. Screenshots attached.

**2. No way to show a puzzle's rules.** A Sudoku, sure, everyone knows
it - but Futoshiki or Shikaku aren't universally familiar, and a print
puzzle book almost always includes a one-line reminder of the rules
somewhere. There was no way to add one.

## What

- The `"puzzle"` source type itself: Sudoku, Binoxxo, Futoshiki, Kakuro, Shikaku generators + `PuzzleStoryProvider` (from #123).
- **Solutions use `PlacementPreference.APPENDIX`**
  instead of `FULLPAGE`. "Batched like a puzzle book's answer
  section" intent, working regardless of
  how many separate puzzle sources a config has.
- **New `explanation` option** on the `"puzzle"` source:
  `"none"` (default) / `"inline"` (a short rules
  blurb repeated under every puzzle instance) / `"footer"` (a real CSS
  footnote - `float: footnote` - at the bottom of whichever page the
  puzzle lands on, not a paragraph glued into the reading flow; every
  instance of that `puzzle_type` in the paper shares one footnote via a
  small reference mark) / `"appendix"` (one rules blurb per `puzzle_type`,
  grouped with the solutions).
- Both `"footer"` and `"appendix"` explanations use a stable per-
  `puzzle_type` headline, so `Goosepaper.get_stories(deduplicate=True)`
  collapses a rules blurb requested by several sources of the same type
  down to one copy - same dedup mechanism #122 already proved works for
  appendix content generally, no new mechanism needed here. Now that #130 is
  merged, `to_html()`/`to_pdf()`/the CLI pass `deduplicate=True` through, so this
  reaches a normal render, not just a direct `get_stories()` call.

## Reproduction

Two sources:
```json
{
  "sources": [
    {"type": "puzzle", "puzzle_type": "sudoku", "difficulty": "medium", "name": "Sudoku"},
    {"type": "puzzle", "puzzle_type": "binoxxo", "difficulty": "hard", "name": "Binoxxo (schwer)"}
  ]
}
```

*Before:*
puzzle solution renders immediately next to its own puzzle (same page, second column).

*After (this branch):*
puzzles are shown together and appendix holds the solutions. Explanations given in footnotes.

Example attached:
[Rätselheft 2026-07-29.pdf](https://github.com/user-attachments/files/30523883/Ratselheft.2026-07-29.pdf)

## Testing
* Full test suite passes (101+ tests, now including #123's puzzle-generation tests directly since it's part of this PR).
* Verified end-to-end against all three real newspaper configs I run
daily - all of them use `explanation: "footer"` with multiple puzzle
sources per issue, exactly the scenario this PR fixes - including a
real delivery run to a reMarkable.

## Merge-order note (siblings, not dependencies)

This PR has no remaining hard dependencies - #122, #123, and #130 are all resolved (merged or absorbed into this PR). What's left is mechanical (non-semantic) overlap with other open PRs that touch the same shared files.

I ran a local pairwise merge check across all my currently open PRs to find real conflicts (not just shared files). Full breakdown and reasoning in my comment below on this PR - short version: this PR conflicts with #133 and #132 (same `config.py` source-type schema block) and with #121 (same block, plus `rss.py`/`test_rss.py` for #121 separately). Suggested landing order for the whole queue is in that comment too.


j6k4m8:
Can you combine this with #123 so all puzzle logic is in the same place? It does make it a bigger PR, but I'd rather that it's not across two PRs. Shouldn't be too hard since #122 is now merged!

Smengerl:
## Combined with #123, and a full merge-order check across all open PRs

**Consolidation:** this PR (`feature/puzzle-explanations`) was branched off #123's branch (`feature/puzzle-provider`) and has been kept in sync with it via merge ever since — `git diff origin/feature/puzzle-provider origin/feature/puzzle-explanations -- goosepaper/puzzlegen` is empty, so this PR already fully contains #123's code. No rebase/squash needed to combine them; I've closed #123 as superseded by this one. This PR is also conflict-free against current `master`.

**Why post an order here at all:** combining two PRs into one changes the shape of the queue, so before doing that I checked all 13 currently-open PRs against each other — not just "do they touch the same file" but an actual local `git merge --no-commit` per pair, to find real conflicts. Posting the full result here since this PR is now the biggest node in that graph; I've left a short pointer comment on each other affected PR linking back to this one.

### Conflict clusters found (real merge conflicts, not just shared files)

**A. `styles.py`** — #120 ↔ #126 (both insert a new CSS block at the same anchor point)

**B. `storyprovider/rss.py` + `test_rss.py`** — #135, #134, #129, #128, #127, #121
Conflicting pairs: (135,134), (135,127), (129,121), (128,127), (128,121), (127,121)

**C. `config.py`/`test_config.py`/`test_utils.py`/`util.py`/`README.md`** (each adds a new source-type registration to the same schema dict) — #133, #132, **this PR**, #121
Every pair conflicts: (133,132), (133,this), (133,121), (132,this), (132,121), (this,121)

#136 also touches `config.py` (`DeliverySettings`) but a different section entirely — zero conflicts with anything, despite the shared file.

**#121 is the hub**: it's in both cluster B and C, conflicting with 6 of the other 12 PRs. Merging it last in both clusters means it absorbs one final rebase instead of forcing repeated conflicts onto its 6 siblings one at a time. This is inherent to the overlapping code (13 conflicting pairs total, independent of ordering) — the order below is about predictability, not eliminating conflicts that can't be eliminated.

### Suggested merge order

1. **#136**, **#131** — no dependencies on anything else, merge whenever
2. **#120 → #126** — styles.py pair
3. **#129 → #134 → #135 → #128 → #127** — rss.py-only PRs
4. **#133 → #132 → this PR (#124)** — config.py source-schema cluster
5. **#121 last** — touches both clusters B and C

Happy to rebase this PR (or any of the others) at whatever point in that order works for you — just say the word.

Smengerl:
**Correction to my comment above:** the "conflict-free against current `master`" line was wrong — I'd tested against my fork's own stale `master` mirror, not `j6k4m8/goosepaper:master` (the actual PR base). Against the real base, GitHub reports this PR as `dirty` (real conflict), caused by #138 (`register_story_provider`, merged 2026-08-08 — not one of my PRs) inserting a test in `test_config.py` at the same spot this PR's own new tests land. Trivial to resolve with a rebase, and unrelated to the sibling-PR conflict clusters/order described above — those were tested branch-to-branch directly, not against master, so they're unaffected by this. I'll rebase this PR once we're ready to act on the merge order.

Smengerl:
Fixed a real rendering bug in Kakuro clue cells: the right-sum ("h") was rendering bottom-right of its clue cell instead of top-right, and the down-sum ("v") top-left instead of bottom-left - backwards from the standard Kakuro convention (right-sum upper-right, down-sum lower-left). Root cause: WeasyPrint inverts the vertical axis for an absolutely-positioned span nested inside a table cell this way (position:relative div inside a position:relative td) - `top: 0` renders at the visual bottom, `bottom: 0` at the visual top. Confirmed by rendering an isolated, distinctly-labelled clue cell before/after. Fix: swap `top`/`bottom` between `.kakuro-h`/`.kakuro-v` (left/right unaffected).

README.md
```diff
@@ -175,6 +175,27 @@ Check out [this example PDF](https://github.com/j6k4m8/goosepaper/blob/master/do
...
+| `count` | optional | `1` | How many puzzle instances of this `puzzle_type`+`difficulty` to generate. |
+| `seed` | optional | random | RNG seed, for reproducible generation. |
+| `explanation` | optional | `"none"` | `"none"`, `"inline"` (a short rules blurb repeated under every puzzle instance), `"footer"` (a real CSS footnote at the bottom of whichever page it lands on - one footnote per `puzzle_type` in the whole paper, with every instance of that type carrying its own small reference mark pointing at it), or `"appendix"` (one rules blurb per `puzzle_type`, grouped with the solutions at the end of the document). |
+| `name` | optional | none | Visible heading for this puzzle instance (and its solution, suffixed " - Lösung"). If omitted, no heading renders for the puzzle itself - only the enclosing section's own title identifies it. Useful when several different puzzle types share one section; redundant (and best left unset) when a section already covers exactly one type+difficulty. |
```
j6k4m8: I don't quite see where this is coming from right now but we should (since the rest of the codebase is currently localized) use english "Solution" / "Answer". I'm not opposed to localization of the repo's included copy text someday but we should at least be consistent about it for now (and internationalize the entire codebase at once if we decide to do so at all).

goosepaper/util.py
```diff
@@ -141,6 +141,23 @@ def construct_story_providers_from_source_configs(source_configs):
             "WikipediaCurrentEventsStoryProvider",
             lambda options: {},
         ),
+        "puzzle": (
```
j6k4m8, line 182: this is a little unfair bc I moved the goalposts after your PR was opened, but would you please add docs for the new story providers per #140; seems like hopefully this is all claude's work burden anyway :)

goosepaper/storyprovider/puzzle.py
```diff
@@ -0,0 +1,557 @@
...
+      - `seed` (optional, default None): RNG seed, for reproducible generation.
+      - `explanation` (optional, default "none"): "none"/"inline"/"footer"/"appendix" - see
+        get_stories()'s docstring for what each mode renders.
+      - `name` (optional, default None): visible per-instance heading - see get_stories()'s
+        docstring for the "no name -> no heading, just the section's own" behavior.
+    """
+
+    def __init__(
+        self,
+        puzzle_type: str,
```
j6k4m8: like the other PR under review right now, probably makes sense to do a `Literal[...]` here to help editors intellisense this

goosepaper/storyprovider/puzzle.py
```diff
@@ -0,0 +1,557 @@
...
+        box_size: int = 3,
+        difficulty: str = sudoku.DEFAULT_DIFFICULTY,
+        count: int = 1,
+        seed: Optional[int] = None,
+        explanation: str = "none",
+        name: Optional[str] = None,
+    ) -> None:
+        if puzzle_type not in _GENERATORS:
```
j6k4m8: this is kinda the same logic as `_validate_puzzle_type` right?

goosepaper/storyprovider/puzzle.py
```diff
@@ -0,0 +1,557 @@
...
+        three difficulties use the same box size, so there is no per-difficulty table for it.
+      - `difficulty` (optional, default "medium"): one of "easy", "medium", "hard". Controls
+        how many cells/constraints are given (see each type's own `DIFFICULTIES` table) and,
+        for every type but sudoku, the grid size itself - see e.g. `binoxxo/config.py`'s
+        `DIFFICULTIES` table for the exact per-difficulty sizes and why grid size and
+        difficulty aren't independent knobs to begin with.
+      - `count` (optional, default 1): how many puzzle instances of this type+difficulty to
+        generate.
+      - `seed` (optional, default None): RNG seed, for reproducible generation.
+      - `explanation` (optional, default "none"): "none"/"inline"/"footer"/"appendix" - see
```
j6k4m8: I think this is probably better named `solution`

j6k4m8 (changes requested):

goosepaper/storyprovider/puzzle.py
Smengerl: `explanation` doesn't show the solution though — the actual solution grid is already always generated and collected in the appendix (see `PlacementPreference.APPENDIX` above), completely independent of this parameter. `explanation` only controls whether/where a short "how does this puzzle type work" rules blurb appears (none/inline/footer/appendix) — so renaming it to `solution` would actually be misleading, since it has nothing to do with the solution itself. I'd like to keep `explanation` as-is unless I'm missing what you had in mind — happy to rename if there's a different angle I'm not seeing.

goosepaper/storyprovider/puzzle.py
Smengerl: Yep, exactly — extracted a shared `_validate_choice()` helper in 6ffb55e, both `puzzle_type` and `explanation` now go through it instead of duplicating the not-in/raise pattern. Error message wording is unchanged.

goosepaper/storyprovider/puzzle.py
Smengerl: Done in 314b6a7 — added `PuzzleType`/`ExplanationMode` `Literal[...]` aliases and typed `puzzle_type`/`explanation` with them, same "kept in sync by hand" comment convention as `ComicType` in #133.

README.md
Smengerl: Fixed in 44fbca3 — README now says " - Solution". Went a step further per your consistency point: the actual provider code had the same hardcoded German (`" - Lösung"` in `puzzle.py`, plus a "Wie funktioniert {type}?" headline), and the `_EXPLANATIONS` rule blurbs shown to readers were German too — all now English, matching the rest of the codebase.

goosepaper/util.py
Smengerl: Added `docs/reference/storyprovider/puzzle.py.md` in 14765e6, following the exact format #140 established (and that #133's `comic.py.md` already used). Also resolved the merge conflict with master and added the README link in 97d1b36 — no worries about the goalposts moving, all done now.

PR #126, opened 2026-08-03
fix: hide interactive <button> elements in print/PDF output

Smengerl:
## What's broken

Article HTML pulled in via RSS/readability sometimes keeps an interactive `<button>` from the
source page — most commonly an image "click to zoom" lightbox trigger, a common
WordPress/Gutenberg pattern. Its real position/visibility is normally set by JavaScript (e.g.
positioned as a small overlay icon in a corner of the image) that never runs in this pipeline,
so the button falls into normal document flow instead, rendered with the UA stylesheet's default
`<button>` styling: grey background, border, rounded corners. Since it typically contains only an
icon — often a light-colored SVG meant to sit on a dark/photo background — it reads as an empty,
flat grey block with nothing legible inside, sitting directly under whatever image it was meant
to overlay.

## Fix

No `<button>` in extracted article prose is ever meaningfully interactive in a static, print/PDF
context, so hide every one uniformly (`display: none`) rather than chase each source site's
specific button/icon class one at a time. Same reasoning already applied here to `<svg>` sizing
(#119, merged) and `<pre>` wrapping (#120, open): defensive, general print-safety CSS rather than
a per-source content filter.

**Overlap note:** this PR and #120 both insert a rule block into `_base_print_css` at the same
point. The rules are independent (`button { display: none; }` vs. the `pre`/`code` wrapping
rules), so no semantic conflict - but merging both will hit a textual conflict at that shared
insertion point; trivial to resolve by keeping both blocks.

## Testing

- Added `test_interactive_buttons_are_hidden` in `test_styles.py`, asserting the base print CSS
  contains `button { display: none; }`.
- Full test suite passes.
- Verified against a real failure case: an Electrek (electric-bike roundup) article with a
  WordPress lightbox-trigger button under every product photo — before the fix, each rendered as
  an empty flat grey box; after, gone entirely, image and caption sit directly against each
  other.

### DIFF
--- a/goosepaper/test_styles.py
+++ b/goosepaper/test_styles.py
@@ -25,6 +25,21 @@ def test_unsized_svg_defaults_to_icon_size():
     assert "width: 1em" in css
 
 
+def test_interactive_buttons_are_hidden():
+    """Article HTML pulled in via RSS/readability sometimes keeps an interactive <button> from
+    the source page - most commonly an image "click to zoom" lightbox trigger (a common
+    WordPress/Gutenberg pattern). Its real position/visibility is set by JavaScript that never
+    runs here, so unpositioned it falls into normal document flow using the UA stylesheet's
+    default <button> styling (grey background, border, rounded corners) - reads as an empty
+    flat grey block, since it typically contains only an icon (often invisible against that same
+    default grey). No <button> in extracted article prose is ever meaningfully interactive in a
+    static, print/PDF context, so every one must be hidden, not just specific known button
+    classes from specific sites."""
+    css = Style("FifthAvenue").get_css(page_profile="paper_pro", layout="2col")
+
+    assert re.search(r"button\s*{\s*display:\s*none;", css)
+
+
 def test_appendix_block_gets_exactly_one_page_break_not_per_entry():
     """The appendix (PlacementPreference.APPENDIX) block must start on a fresh page - but only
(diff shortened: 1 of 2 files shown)


j6k4m8:
This is smart — we should survey briefly that we're not losing things like (poorly-constructed) `<button>` used as a link or other antipatterns which would break fluency. But this is otherwise a great addition, will merge shortly!

goosepaper/test_styles.py
```diff
@@ -25,6 +25,21 @@ def test_unsized_svg_defaults_to_icon_size():
     assert "width: 1em" in css
 
 
+def test_interactive_buttons_are_hidden():
```
j6k4m8: Like #120 I don't think we need CSS that just checks that hardcoded strings that we set are still set; let's kill this test since it's not really accomplishing anything besides increasing CI test count. Otherwise LGTM!

j6k4m8 (changes requested):

Smengerl:
Dropped the brittle string-assertion test as requested — the CSS diff will make it obvious if the button-hiding rule gets reverted. Pushed in 22a3cf5.

Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #120 — both insert a new CSS block at the same anchor point in `styles.py`'s `_base_print_css`. Suggested order: after #120.

j6k4m8 (approved):

PR #127, opened 2026-08-03
fix: fall back to sniffed encoding when a page omits an explicit charset

Smengerl:
## Summary
- `requests` defaults undeclared `text/*` charsets to ISO-8859-1 per RFC 2616. When a fetched article page's `Content-Type` header doesn't declare a charset, this mangles UTF-8 pages (accented characters etc. get decoded as the wrong encoding).
- Fall back to `response.apparent_encoding` (requests' own content-sniffing) in that case, before the body is read - so `_story_from_response` sees correctly decoded text.
- No behavior change for pages that do declare a charset in their header.

## Test plan
- [x] `pytest goosepaper/storyprovider/test_rss.py` - 11 passed (9 existing + 2 new: an undeclared charset gets sniffed and decoded correctly; a declared charset is left untouched, even when sniffing would've suggested something different)

## Merge overlap note

Verified by locally merging every pairwise combination of my currently open PRs against `master`. This one produces a mechanical (non-semantic) merge conflict with **#121, #128, #135** - all touch `rss.py`/`test_rss.py` at nearby insertion points. No logic overlap in any case checked.

### DIFF
--- a/goosepaper/storyprovider/rss.py
+++ b/goosepaper/storyprovider/rss.py
@@ -154,6 +154,13 @@ def _story_from_response(
     fallback_body_html: str = "",
     prefer_feed_title: bool = False,
 ) -> Story:
+    content_type = response.headers.get("content-type", "")
+    if "charset" not in content_type.lower():
+        # Per RFC 2616, requests defaults undeclared text/* charsets to
+        # ISO-8859-1, mangling UTF-8 pages that omit an explicit charset.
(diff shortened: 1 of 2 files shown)


Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #135, #128, and #121 — the most contested spot in `storyprovider/rss.py`'s `_story_from_response`. Suggested order: last among the rss.py-only group, after #129 → #134 → #135 → #128.

PR #128, opened 2026-08-03
fix: don't drop an entire feed's stories when one entry fails to fetch

Smengerl:
## Summary
- `RSSFeedStoryProvider.get_stories()` had no per-entry error handling: a single broken link (e.g. a feed linking out to a dead or SSL-broken third-party site) raised out of `_story_from_entry`, and since that call sat outside any try/except, the exception propagated out of `get_stories()` itself - dropping every other story from that source for the run.
- Split `_story_from_entry` into a thin wrapper that catches and logs, and `_story_from_entry_unsafe` holding the original logic unchanged. One bad entry is now skipped instead of taking down the whole feed's batch.

## Test plan
- [x] `pytest goosepaper/storyprovider/test_rss.py` - 10 passed (9 existing + 1 new: two entries, one raises `ConnectionError`, the other still comes through)
- [x] Full suite (`pytest`) - 79 passed

## Merge overlap note

Verified by locally merging every pairwise combination of my currently open PRs against `master`. This one produces a mechanical (non-semantic) merge conflict with **#121, #127** - both add a new test to `test_rss.py` near where this PR's own test lands. No conflict in the actual `rss.py` implementation in either case.


j6k4m8:
Thank you @Smengerl — feeling a little underwater. Can you give me your preferred review / merge order for all these PRs?

Smengerl:
Totally understandable — that's a lot of PRs to land on at once, sorry about the pile-up! I went through all of them again, checked actual dependencies and simulated the pairwise merges locally to see where real conflicts (vs. just adjacent-line noise) would land. Here's my suggested order:

**Phase 1 — independent, no overlap with anything else, safe to merge in any order:**
1. #130 (deduplicate param)
2. #131 (PDF bookmark levels) — confirmed conflict-free with #130
3. #120 (code block wrapping) — CI had gone red in the meantime, just fixed
4. #126 (hide interactive buttons in print) — only touches the same file as #120 (`styles.py`) at one shared insertion point; purely textual, no logic overlap
5. #136 (delivery retention)
6. #137 (interactive auth flag) — merging both #136 and #137 locally surfaced one real (non-textual) test conflict between them; already fixed on #136's side, so order between the two no longer matters

**Phase 2 — puzzle pair, hard dependency:**
7. #123 (puzzle provider)
8. #124 (puzzle explanations) — depends on #123 (+ already-merged #122); verified this merges cleanly and isolates to exactly the 6 files/±diff described in its own PR body

**Phase 3 — more source types, share the config.py/util.py registries:**
9. #133 (comic provider)
10. #132 (section provider) — also touches `_parse_source` itself, but nothing else currently open touches those specific lines, so it doesn't add extra conflict surface

**Phase 4 — RSS fixes, share rss.py:**
11. #128 (per-entry error handling)
12. #129 (prefer feed title)
13. #127 (encoding fallback)
14. #134 (duplicate embedded headline)
15. #135 (relative URLs) — conflicts with #134 at the same insertion point in `_story_from_response` (textual only, the two are logically independent, matches what #135 itself says)

**Last — #121 (RSS content filters):**
It overlaps with 7 of the other 15 PRs, more than any other. Merging it last means only #121 itself needs a (bigger) rebase, instead of 7 other PRs each needing a small one against it. Trade-off worth flagging: this delays #121 landing — if you'd rather have it available sooner, it can move earlier instead, at the cost of ~7 small follow-up rebases on my end.

Every conflict I flagged above is textual (adjacent insertions in the same function/registry) rather than semantic — happy to handle the rebases on my side as we go through this, just let me know the pace that works for you.

Also, in case it's reassuring rather than alarming: this is (for now) the last batch of PRs from me — I'm pretty happy with the newspapers I'm getting out of this these days, so my todo list of "things goosepaper should also do" has finally run dry. And if you're curious what it's for: I wrapped this into a Home Assistant add-on so it generates and delivers a fresh edition to my reMarkable every morning without needing a separate server: https://github.com/Smengerl/goosepaper-addon

Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #127 and #121 — same region of `storyprovider/rss.py`. Suggested order: #129 → #134 → #135 → this PR → #127.

j6k4m8:
@Smengerl I think a few of these PRs have gone slightly stale as I merge your contributions in; would you be willing to update them with latest trunk and make sure tests pass etc? Would also appreciate fresh guidance on which order you'd like them reviewed. I'm going as fast as I can but this is all human review, not LLM :)

PR #129, opened 2026-08-03
feat: optionally prefer the feed's own title over readability's extracted title

Smengerl:
## Summary
- readability's `doc.title()` is unreliable on some sites (e.g. it returns just the site name for every article on some blogs). The RSS feed's own `<title>` is usually accurate.
- New `prefer_feed_title` flag, off by default so existing behavior is unchanged; callers who hit this on a specific feed can opt in per source.

## Test plan
- [x] `pytest goosepaper/storyprovider/test_rss.py` - 11 passed (9 existing + 2 new: flag on -> feed title wins; flag off/default -> readability title unchanged)

## Merge overlap note

Verified by locally merging every pairwise combination of my currently open PRs against `master`. This one produces a mechanical (non-semantic) merge conflict with **#121** only - both add new parameters to `RSSFeedStoryProvider.__init__` at the same point. No overlap with any other currently open PR.

### DIFF
--- a/goosepaper/storyprovider/rss.py
+++ b/goosepaper/storyprovider/rss.py
@@ -22,6 +22,7 @@ class RSSFeedStoryProvider(StoryProvider):
         since_days_ago: int = None,
         byline: str = "all",
         body_source: str = "auto",
+        prefer_feed_title: bool = False,
     ) -> None:
         if byline not in RSS_BYLINE_MODES:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #121 (both touch `storyprovider/rss.py`; #121 also separately conflicts with the config.py cluster — see #124). Suggested order: this PR first among the rss.py group → #134 → #135 → #128 → #127, with #121 merged last overall.

j6k4m8 (approved):
I am somewhat ambivalent about these tests, they're meaningfully more sophisticated than the actual code they test and I worry they're not quite usefully testing the behavior we care about. I'm cool with leaving them in but I'm leaving this review note so that in the future if we encounter issues with them we can feel less bad about taking them back out :)

PR #130, opened 2026-08-03
fix: actually expose the existing deduplicate option to callers

Smengerl:
## Summary
- `Goosepaper.get_stories(deduplicate=...)` already existed, but nothing else ever passed a value through: `_render_html_document()` (used by both `to_html()` and `to_pdf()`) calls `self.get_stories()` with no arguments, so the option was unreachable except by calling `get_stories()` directly.
- Add a `deduplicate` constructor argument that `get_stories()` now falls back to when its own parameter is omitted, so `to_html()`/`to_pdf()` honor it too.
- Default unchanged (`False`), so existing behavior is untouched unless a caller opts in.

## Why this matters beyond this PR

This closes a gap that #124 (puzzle-explanations) currently has: #124's `"footer"`/`"appendix"` explanation modes rely on `Goosepaper.get_stories(deduplicate=True)` to collapse a rules blurb requested by several sources of the same puzzle type down to one copy. #124's own tests call `get_stories(deduplicate=True)` directly to verify this, but neither `goosepaper/__main__.py` nor `to_html()`/`to_pdf()` pass `deduplicate=True` through today - so a real user generating a paper via the CLI/Docker wouldn't actually get that dedup behavior, only a caller that bypasses `to_html()`/`to_pdf()` and calls `get_stories()` directly would. This PR is what makes that dedup reachable from a real render, e.g. `Goosepaper(providers, deduplicate=True).to_pdf(...)`.

## Test plan
- [x] `pytest goosepaper/test_goosepaper.py` - 15 passed (13 existing + 2 new: `to_html()` deduplicates when the constructor flag is set; doesn't by default)

### DIFF
--- a/goosepaper/goosepaper.py
+++ b/goosepaper/goosepaper.py
@@ -37,6 +37,7 @@ class Goosepaper:
         story_providers: List[StoryProvider],
         title: str = None,
         subtitle: str = None,
+        deduplicate: bool = False,
     ):
         """
         Create a new Goosepaper.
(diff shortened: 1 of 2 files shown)


j6k4m8 (approved):
Nice, good fix!

PR #132, opened 2026-08-03
feat: add SectionProvider wrapper for tagging any provider's stories with a section

Smengerl:
## Summary
- Goosepaper already renders stories grouped by `Story.section_title` (`_story_runs`/`_render_story_region`), but nothing in the library sets that field - callers assembling a paper out of several providers had no generic way to assign each one to a named section.
- `SectionProvider` wraps any `StoryProvider` and tags its output with a section title.

## Test plan
- [x] `pytest goosepaper/storyprovider/test_section.py` - 2 passed (section tag gets set; headline left unchanged)

## Edit history

An earlier version of this PR also gave `SectionProvider` an optional `headline_prefix`
(prepended to every headline, e.g. so two same-typed providers - "two weather providers for two
different locations" - could still be told apart). Checked it against every real newspaper
config I run and every bundled example: none use more than one instance of the same provider
type, so nothing needs it, and the JSON config schema never exposed it either (Python-API-only).
Dropped it to keep this PR scoped to what `section` itself is actually for.

## Merge overlap note

Verified by locally merging every pairwise combination of my currently open PRs against `master`. This one produces a mechanical (non-semantic) merge conflict with **#121, #123, #124, #133** - all touch the same README/example-config/test areas this PR's `section` field also documents. `config.py`/`util.py` themselves merge clean against all of them - this PR is the only one that changes `_parse_source`'s/`_source_config_parts`'s own contract, and nothing else currently open touches those specific lines.


Smengerl:
Find example files before/after PR attached:
[with_sections.pdf](https://github.com/user-attachments/files/30668853/with_sections.pdf)
[without_sections.pdf](https://github.com/user-attachments/files/30668854/without_sections.pdf)

Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #133, #124 (the puzzle PR), and #121 — same `config.py` source-schema block. Suggested order: #133 → this PR → #124 → #121 last overall.

Smengerl:
Heads-up, separate from the sibling-conflict note above: GitHub currently reports this PR as `dirty` against `master` — caused by #138 (`register_story_provider`, merged 2026-08-08, not one of my PRs) touching `test_config.py`/`util.py` at the same spot this PR's own additions land. Trivial rebase, unrelated to the #132/#133/#124/#121 conflict cluster.

README.md
```diff
@@ -141,6 +141,10 @@ The paper config uses a strict v2 schema with one file per paper. A minimal exam
...
 
+Any source can also set `"section": "Tech"` to render its stories grouped under that heading, alongside every other source sharing the same section name - useful once a paper mixes several feeds and you want them organized into named groups rather than one flat run of stories. A source with no `"section"` renders ungrouped, same as today.
```
j6k4m8: Does this mean that you would have `section_heading_visible` in an _item_ in the config? I don't quite get the semantics of that,

```json
{
    "type": "news1",
    "section": "News",
    "section_heading_visible": false, // ←
},
{
    "type": "news2",
    "section": "News",
    "section_heading_visible": true, // what happens here?
},
```

Smengerl:
Good question — I traced it and pinned it with tests.

**In your example (both items listed consecutively):** they form one `News` run. `section_heading_visible` is resolved per run, and hiding is the sticky outcome — `false` on either item wins, the `true` on the other has no effect. So the `News` heading is hidden (still a jump target in the TOC). Order of the two items doesn't matter. This mirrors the existing `test_toc_hides_section_heading_if_any_story_in_the_run_wants_it_hidden`, just via config rather than raw `Story` objects — `_render_story_region` does `if not all(story.section_heading_visible for story in run_stories)`.

**The one sharp edge:** grouping is by *consecutive* run (`_story_runs`), not global. If a different section sits between two `News` items, they split into two runs and the `News` heading renders twice — once hidden, once visible (`section-news` / `section-news-2`).

I went with documenting that as a constraint rather than making `_story_runs` group globally — global grouping would silently reorder stories across a scattered section, and "list the sources for one section together" is a reasonable ask. Pushed 2e247fb: README + `SourceConfig.section` note, plus tests for both the supported case and the split-run behavior. Also merged current master (the #133 conflict was just the adjacent new kwarg in `Story.__init__` and a new test in `test_utils.py`).

PR #133, opened 2026-08-03
feat: add "comic" source type for daily comic strips (XKCD, gocomics.com, arcamax.com)

Smengerl:
## Motivation

A comics section is a staple of just about every physical newspaper - and every developer loves
XKCD. This adds a native `"comic"` source type so a goosepaper edition can include one, without
needing a screen-scraping RSS-hack workaround.

## What it does

Downloads today's strip and embeds it as a single image story:

```json
{ "type": "comic", "comic_type": "xkcd" }
```

```json
{ "type": "comic", "comic_type": "gocomics", "comic_name": "calvinandhobbes" }
```

| Option | Required? | Default | What it does |
|---|---|---|---|
| `comic_type` | **required** | - | One of `xkcd`, `gocomics`, `arcamax`. No default on purpose - a config that forgets it fails loudly instead of silently always fetching XKCD. |
| `comic_name` | required for `gocomics`/`arcamax`, rejected for `xkcd` | - | The comic's own slug on that site, exactly as it appears in the site's URL - e.g. `"garfield"` or `"calvinandhobbes"` for gocomics.com, `"beetlebailey"` for arcamax.com. |

`gocomics.com` and `arcamax.com` each serve *every* comic in their catalog (hundreds of strips)
through one identical URL/markup template - only the slug differs. So rather than one hardcoded
`comic_type` per specific comic, those two are generic site-backed sources: any comic hosted on
either site works just by passing its slug as `comic_name`, verified live against several
different strips on each site. `xkcd` only ever serves one comic and takes no `comic_name`.

The headline is a fixed, source-derived name, never the strip's own per-day title, and no byline
is set - a byline or a per-day dynamic headline would just repeat the same source name the
headline already shows, or produce a needlessly long heading, unlike a byline on an RSS article
(which distinguishes otherwise-anonymous entries pulled from different feeds). For `xkcd` it's
the fixed string `"XKCD"`; its real per-day title is still available via the embedded image's
`alt` attribute, and its mouseover joke still renders as a caption underneath. For
`gocomics`/`arcamax` - where one source entry covers hundreds of possible comics, so there's no
per-comic label to hardcode - the headline is instead read straight off the fetched page itself:
gocomics.com's structured `ComicSeries` JSON-LD, arcamax.com's `og:title` meta tag.

gocomics.com's page URL is date-scoped (`.../<comic_name>/YYYY/MM/DD`). Its daily
publish-rollover time/timezone isn't documented, so a generation run earlier in the day than
that rollover would otherwise fail outright every time; rather than guess at a timezone, a miss
now retries a few days back automatically before giving up; a log line notes it, nothing shows
up differently in the newspaper itself.

The strip image is downloaded, decoded, and re-encoded (not linked by remote URL and not passed
through unmodified) via Pillow:
- Resolves an image-embedding bug where WeasyPrint silently dropped the *entire* story - no
  exception, no log line - for certain source images (a CDN serving print resolution up to
  2800px wide with no smaller variant, a lossless PNG re-encode of gradient-heavy color art being
  several times larger than the same content as JPEG, and CMYK-mode JPEGs with large embedded
  Photoshop/ICC metadata all independently triggered it). Bounding pixel dimensions, normalizing
  color mode, and using JPEG's lossy compression instead of PNG fixes all three.
- Composites any transparency onto white before dropping the alpha channel, rather than leaving
  Pillow's default behavior (keep whatever RGB value was stored under a transparent pixel,
  unchanged) show phantom colors - not observed against any real source tested so far, but a
  real gap for a hypothetical source that does use transparency.
- Inlining as a base64 `data:` URI (rather than a remote ``) also sidesteps gocomics.com
  requiring the same browser-like headers for the image request as for the page request -
  WeasyPrint fetches `` URLs itself while rendering the PDF, with no way to attach them.

## Demo

Rendered (with the original `xkcd`/`cah`/`garfield` version of this PR, before the slug-generic
rework above - the embedding/rendering pipeline itself is unchanged) across a few different
bundled styles/layouts/page profiles, to show it holds up across genuinely different goosepaper
configurations - not just one specific theme:

- **Academy, 1 column, `remarkable2`** - a compact single-column e-reader layout.
- **FifthAvenue, 2 columns, `paper_pro`** - side-by-side comics reading like adjacent newspaper
  boxes.
- **Autumn, 3 columns, `a4`** - a dense, print-style multi-column page.

Happy to paste screenshots in here directly if useful - just say the word.

## Testing

- 105 tests pass (`uv run pytest`), including dedicated regression tests for: the
  image-embedding bug and its three contributing causes, the transparency-compositing fix, an
  unknown `comic_type` being rejected, `comic_name` being required/rejected per `comic_type`,
  headline derivation from each site's own page markup, and the gocomics.com retry-on-miss
  behavior.
- Verified live against all three real sites (xkcd.com, gocomics.com, arcamax.com), across
  several different comic slugs on gocomics.com/arcamax.com, as of this writing.

## Merge overlap note

Verified by locally merging every pairwise combination of my currently open PRs against `master`. This one produces a mechanical (non-semantic) merge conflict with **#121, #123, #124, #132** - all add a new source type/fields to the same shared registries in `config.py`/`util.py`. No logic overlap in any case checked.


Smengerl:
image-1785776075705
image-1785776081164
image-1785776086835

Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #132, #124 (the puzzle PR), and #121 — all add a new source-type entry to the same `config.py` schema block. Suggested order: this PR first in that group → #132 → #124 → #121 last overall.

goosepaper/storyprovider/comic.py
```diff
@@ -0,0 +1,225 @@
...
+    alone was not enough to bring the payload down to a safe size; (3) arcamax.com's Garfield
+    JPEGs ship CMYK-mode pixel data with a large embedded Photoshop/ICC metadata block. Any of
+    these, combined with the hundreds of other images already in a full newspaper, was enough to
+    trigger the failure. Re-encoding through Pillow bounds the pixel dimensions, normalizes color
+    mode (e.g. CMYK -> RGB), and uses JPEG's lossy DCT compression - far more compact than PNG for
+    this kind of photo-like, gradient-heavy content - regardless of what the source serves on a
+    given day.
+    """
+
+    def __init__(self, comic_type: str, date: Optional[datetime.date] = None) -> None:
```
j6k4m8: we could perhaps type `comic_type` as `Literal["xkcd", "cah", "garfield"]` or (idk if this works? untested) `Literal[_ComicSource.keys()]` or something, so that editors can help intellisense/hint this

pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
   "ebooklib",
   "readability-lxml",
   "remarkapy>=0.2.2,<0.3",
+  "pillow",
```
j6k4m8, line 32: this is a pretty big dep to add... if there's any way to not have to do image conversions inline in comic downloads it would be super awesome, since `pillow` will nearly double our installed package size (the other main offender is weasyprint but there's simply no way around that one). Did you experiment with other techniques, is this really the lowest-resistance path? Comics are a commonly requested storyprovider so this may be necessary...

j6k4m8: huh, apparently we already have pillow as a deep dependency from some other package. bummer! but also, nice, this means we can ignore this bundle size I guess :)

Smengerl:
Update: reworked this to be much more generic. Instead of one hardcoded `comic_type` per specific comic (`cah`, `garfield`), `gocomics`/`arcamax` are now generic site-backed sources that take any comic's own slug via a new `comic_name` field (e.g. `{"comic_type": "gocomics", "comic_name": "beetlebailey"}`). Both sites serve their entire catalog (hundreds of strips) through one identical URL/markup template, so this unlocks any comic on either site with no code change - verified live against several strips per site. Headline is now derived straight from the fetched page (gocomics.com's JSON-LD, arcamax.com's `og:title`) instead of a hardcoded per-comic label. `xkcd` is unchanged.

Smengerl:
Added reliability fix for gocomics.com: if a comic's strip for the requested day isn't published yet, it now silently falls back to the most recent day that *is* available (up to a few days back) rather than dropping the section. The reader may briefly see a repeat of an older strip instead of that day's - only a log line marks this, nothing shows up differently in the newspaper itself.

goosepaper/storyprovider/comic.py
Smengerl: Done - `comic_type` is now `Literal["xkcd", "gocomics", "arcamax"]` (values updated for the generic-by-slug redesign above; `Literal[_ComicSource.keys()]` doesn't work, so it's a manually-kept-in-sync alias next to `_COMIC_SOURCES` instead). Runtime check stays too, since config-driven callers pass a plain str.

Smengerl:
Update: extracted the Pillow re-encode step (bound dimensions, normalize color mode, composite transparency, always JPEG) out of this file into a new `storyprovider/imageutil.py`, shared with a sibling fix (RSS-sourced article images need the exact same defensive treatment, for the exact same WeasyPrint-silently-drops-the-story reason - not yet a PR here, but landing the same way). No behavior change here, just factoring out logic that's no longer comic-specific. Both PRs carry their own copy of `imageutil.py` so either can merge first without depending on the other.

Smengerl:
The sibling fix mentioned above is now #142.

goosepaper/storyprovider/comic.py
```diff
@@ -183,24 +181,20 @@ class DailyComicStoryProvider(StoryProvider):
...
-    strip is itself several times larger than the same content as JPEG - re-encoding as PNG
-    alone was not enough to bring the payload down to a safe size; (3) arcamax.com's Garfield
-    JPEGs ship CMYK-mode pixel data with a large embedded Photoshop/ICC metadata block. Any of
-    these, combined with the hundreds of other images already in a full newspaper, was enough to
-    trigger the failure. Re-encoding through Pillow bounds the pixel dimensions, normalizes color
-    mode (e.g. CMYK -> RGB), and uses JPEG's lossy DCT compression - far more compact than PNG for
-    this kind of photo-like, gradient-heavy content - regardless of what the source serves on a
-    given day.
+    via `imageutil.reencode_image_as_data_uri` - the same "decode, then adjust, then re-encode"
+    pipeline remarkable_news's own Go tool runs (imaging.Decode + resize + imaging.Save with
```
j6k4m8, line 190: what's `remarkable_news`? is that an artifact of prompting?

Smengerl: `evidlo/remarkable_news` is a separate, independent open-source project (unrelated to goosepaper or my own addon) — [github.com/Evidlo/remarkable_news](https://github.com/Evidlo/remarkable_news), 299 stars, actively maintained, GPL-3.0: it renders daily comics/news onto a reMarkable tablet's suspend screen via a small Go CLI. What I ported from it is just the *pattern* its per-comic systemd services use (`services/xkcd.service`, `services/cah.service`, `services/garfield.service`) — the page URL, HTTP headers, and XPath needed to locate each site's strip image. Reimplemented natively in Python here instead of shelling out to their CLI.

PR #134, opened 2026-08-04
Work around a headline duplicated inside readability's extracted article body

Smengerl:
Small, fairly narrow fix — I won't be offended if you'd rather not carry this kind of heuristic in the core library. Explained below, happy to adjust or close.

## The problem

A few sites embed the article's headline as a heading (`<h1>`/`<h2>`/`<h3>`) *inside* the same container `readability` identifies as "the article body", instead of keeping the page title and content structurally separate. `readability` has no way to know that heading duplicates the page title rather than being real content, so it stays in `doc.summary()`'s output — the rendered story then shows the headline twice.

Confirmed on three well-known, unrelated feeds (not a one-off):

- **Engadget** (`https://www.engadget.com/rss.xml`) — 4/4 sampled articles affected. Example: [this article](https://www.engadget.com/2230464/we-just-got-our-first-real-look-at-how-small-x-ad-business-is/) — `<h1 class="title-gallery">` nested a couple of `<div>`s deep inside `<div class="news-article">`.
- **The Register** (`https://www.theregister.com/headlines.atom`) — 3/4 affected. Example: [this article](https://www.theregister.com/ai-and-ml/2026/08/04/openai-wants-teachers-and-profs-to-foist-their-work-off-on-chatgpt/5283064) — the heading is preceded by a short `<p class="kicker">` category label ("ai and ml").
- **MacRumors** (`https://feeds.macrumors.com/MacRumors-All`) — 5/7 individual articles affected (not roundup/listicle pages). Example: [this article](https://www.macrumors.com/2026/08/04/telegram-extortionist-app-store-removal/) — `doc.title()` here also carries MacRumors' own `" - MacRumors"` `<title>` suffix, which the embedded heading itself doesn't have.

I also sampled 24 other well-known feeds (BBC, CNN, NYT, The Guardian, Reuters, The Verge, Ars Technica, TechCrunch, Wired, heise.de, and more) and found no trace of this — so it's real, but seems to be specific to how a handful of sites structure their article markup, not something every feed hits.

## Before / after

Top of the rendered page, same three real articles, with vs. without the fix:

**Engadget**

Engadget before
Engadget after

**The Register** (note the short "ai and ml" kicker correctly survives — only the actual duplicate heading is removed)

The Register before
The Register after

**MacRumors** (the `" - MacRumors"` title suffix case)

MacRumors before
MacRumors after

## The fix

Admittedly a bit of a heuristic (`_strip_duplicate_leading_heading` in `rss.py`): it walks the parsed body looking for a *leading* `h1`/`h2`/`h3` — tolerating a short kicker/label before it, and a common `" - Sitename"` title-suffix mismatch — whose text matches the resolved headline, and drops just that one element. It only ever touches the very first heading reached after little preceding text; a heading deeper in genuine content, even one that happens to repeat the headline verbatim, is left alone. I tried to stay conservative rather than clever, since I'd rather under-fix than accidentally eat real content on some feed I haven't tested.

## Testing

8 new tests (7 unit tests on the helper in isolation, covering the nested-wrapper/kicker/title-suffix/non-matching/deeper-heading cases individually, + 1 integration test through `RSSFeedStoryProvider`). Full suite passing. Also manually verified end-to-end against the three real feeds above (not just fixtures) — screenshots in this description are the actual output.

---

**Edit:** pushed one more small commit - found while touching this code again for an unrelated fix, so figured it belonged here rather than a separate PR. `_strip_duplicate_leading_heading` was re-serializing via `str(soup.body)`, and bs4's lxml parser always wraps a bare fragment in a synthetic `<html><body>` internally - every story it actually modified came back with a stray literal `<body>` tag it never had going in. `container.decode_contents()` instead of `str(container)` fixes it; added a regression test that checks exact output rather than just substrings (which is why the existing tests hadn't caught it).

## Merge overlap note

Verified by locally merging against `master` alongside my other currently open PRs. This one produces a mechanical (non-semantic) merge conflict with **#135** - both insert their own post-processing step into `_story_from_response` right after `doc.summary()` extraction. The two operate independently (element removal vs. attribute rewriting) - resolving means keeping both new lines/helper functions, order doesn't matter functionally.


Smengerl:
**Follow-up fix pushed to this branch** (d1bc037): further production use surfaced a real-world case this original fix didn't cover.

`_strip_duplicate_leading_heading()` only ever recognized `h1`/`h2`/`h3` as a duplicate-heading candidate. Confirmed live against **heise.de** (all three heise sources - heise-atom.xml, heise/ix, heise/developer): heise auto-generates a table-of-contents widget at the very top of every article body:

```html
<a-collapse class="a-toc">
  <nav><ol class="a-toc__list">
    <li class="a-toc__item a-toc__item--current">
      <span aria-current="page" class="a-toc__text">Amazon plant gigantisches Gaskraftwerk für neues KI-Rechenzentrum</span>
    </li>
  </ol></nav>
</a-collapse>
```

The current page's own TOC entry holds the plain headline text with **no heading tag at all** - a `<span>` inside an `<li>`. Two compounding bugs let it through untouched:

1. The tag-name check (`node.name not in ("h1", "h2", "h3")`) never even considered it a candidate.
2. Because of that, its ~60+ characters fell into the generic "preceding chrome" text counter and blew the 40-char budget on their own, before the function had any chance to recognize a duplicate - so the whole function bailed out even for articles where a real duplicate heading *would* have appeared shortly after.

Result in production: headline rendered twice in the PDF, with a spurious list bullet/number in front of the second copy (confirmed on two different live heise.de articles).

**Fix:** check every `Tag`'s full text against the headline as soon as it's encountered, not just h1/h2/h3. Since `.descendants` visits a tag before its own text (a `<span>`'s text is a separate, later descendant of its parent), matching at the tag level and returning immediately means a matched element's own text never reaches the chrome-budget counter - only genuinely unrelated preceding text counts against it now. Matching at the tag level also means the *outermost* matching wrapper gets removed (the whole `<a-collapse>` widget, not just the innermost `<span>`) - stripping only the inner text node would've left an empty `<li>` whose bullet/number would still show up in the PDF.

Minor, strictly-positive side effect: no longer bails out unconditionally on the first non-matching `h1`/`h2`/`h3` - a short non-matching leading heading no longer blocks a real duplicate found just after it, as long as the chrome budget isn't exceeded. The budget check remains the only guard against touching a duplicate buried in genuine body content.

3 new tests added (the heise TOC-widget case, the short-non-matching-heading case, plus manual end-to-end verification against the real heise.de HTML that triggered this in production). Full suite green, no regressions in the existing Engadget/Register/MacRumors h1-h3 cases.

Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #135 — both touch the same region of `storyprovider/rss.py`'s `_story_from_response`. Suggested order: #129 → this PR → #135 → #128 → #127.

j6k4m8 (changes requested):
This is for sure a little delicate, but, well, so is a lot of this codebase :) And this is an obvious quality-of-life improvement for a lot of sites! 

Can you remove the unnecessary screenshots in `docs/`? Don't need those to be permanent parts of the repo. Otherwise looks good to merge to me!

Smengerl:
Good catch — sorry, these shouldn't have made it into the repo in the first place. Removed in f2c9b83.

j6k4m8 (approved):

PR #135, opened 2026-08-05
fix(rss): resolve relative URLs in readability-extracted article bodies

Smengerl:
## What

Absolutizes relative URLs (``, `<source src>`, `<a href>`) in the HTML body that `readability-lxml` extracts from a fetched article, before it becomes a `Story`'s `body_html`.

## Why

`readability.Document.summary()` returns the extracted article body more or less as-is from the source page's own markup. That markup can (and often does) use relative URLs - e.g. `` or a page-relative `<a href="../other-story/">`.

goosepaper renders the whole newspaper - every story from every source, concatenated - as a single HTML document with one `base_url` for the entire thing (`Goosepaper.to_pdf`, which sets it to the local filesystem's `cwd`). There's no single correct base for a multi-origin document, so a relative URL that survives into `body_html` resolves against the wrong thing and silently fails. The most visible symptom is a missing image and a WeasyPrint log line like:

```
Failed to load image at 'file:///assets/img/foo.svg': ... No such file or directory
```

## Fix

Right after `doc.summary()` extracts the body, absolutize `src`/`href` on ``, `<source>`, and `<a>` tags against the article's own final URL (`response.url`, i.e. after any redirects) - the one point where a single correct base is actually known, for that specific story. Already-absolute URLs, protocol-relative URLs (`//cdn.example.com/...`), and `data:` URIs are left untouched.

## Evidence

Reproduced live against current, publicly reachable caranddriver.com review pages (fetched via the RSS feed, same code path this fix touches):

- https://www.caranddriver.com/reviews/a73298183/2026-lexus-rx-350h-hybrid-f-sport-test/
- https://www.caranddriver.com/reviews/a73297467/2027-kia-telluride-x-pro-test/

Both extract body markup containing root-relative `src`/`href` values (site design-token icon SVGs, author profile links). Before the fix these are left relative in `body_html`; after the fix they resolve to absolute `https://www.caranddriver.com/...` URLs, and nothing relative is left behind.

## Testing

- New `TestMakeUrlsAbsolute` test class (7 cases): root-relative image `src`, page-relative link `href`, already-absolute URLs left alone, `data:` URIs left alone, no synthetic `<body>` wrapper leaked into the result (bs4's `lxml` parser always wraps a bare fragment in `<html><body>`, so the implementation uses `.decode_contents()` rather than `str()` on the container), a true no-op returns the original string object unchanged, and empty/missing input handled.
- Full existing test suite passes (85/85), including the `_FakeResponse` fixture update needed to give it a realistic `.url` now that `_story_from_response` reads it.

## Scope note

This is deliberately independent of #134 (duplicate embedded headline) - different bug, different code path, no shared logic - so it's a separate PR rather than bundled in.

## Merge overlap note

Verified by locally merging against `master` alongside my other currently open PRs. This one produces a mechanical (non-semantic) merge conflict with **#127, #134** - all touch `rss.py`/`test_rss.py` at nearby points (#134 at the exact same insertion line in `_story_from_response`, see its own description). No logic overlap in any case checked.


Smengerl:
**Follow-up fix pushed to this branch** (15e71f1): while this PR was still open, further production use surfaced a bug in the original fix.

`_make_urls_absolute()` skipped any URL where `urllib.parse.urlparse(value).netloc` was truthy, intending to detect already-absolute URLs and leave them alone. Protocol-relative URLs (``) also parse with a non-empty `netloc` but no `scheme`, so they were mistakenly treated as already-absolute and left untouched. They then got resolved later against the document's `file://` base_url instead of the article's own URL, producing broken image sources like:

```
Failed to load image at 'file://images.example.com/foo.jpg': URLError: ... No such file or directory
```

Fix: check `.scheme` instead of `.netloc` - that's the only thing that actually distinguishes a truly-absolute URL from a protocol-relative one. Added a regression test (`test_resolves_a_protocol_relative_image_src`) covering this case. Full suite still green.

Pushing straight to this branch rather than opening a new PR since it's the same underlying fix/function and this one's still open - happy to split it out separately if that's easier to review.

Smengerl:
Fixed the flake8 F811 failure (7bcbdf2): my manual conflict resolution for the previous push left a duplicate \`TestMakeUrlsAbsolute\` class further down in the file. Removed the redundant copy - both flake8 checks and the full suite (86/86) are green locally now.

Smengerl:
Part of a merge-order check across all my open PRs (full breakdown on #124). This PR conflicts with #134 and #127 — same region of `storyprovider/rss.py`'s `_story_from_response`. Suggested order: #129 → #134 → this PR → #128 → #127.

PR #137, opened 2026-08-05
feat: add interactive parameter to auth_client()/upload()

Smengerl:
## What

Adds `interactive: bool = True` to `auth_client()` and `upload()`, threaded through to a new `--no-interactive` flag on the standalone `upload_to_remarkable` CLI.

## Why

`auth_client()` builds its `Client` with remarkapy's own `interactive=True` default. That's the right default on a real terminal - a missing device token falls into an `input()`-based pairing wizard. But in any unattended context (a cron job invoking `--deliver` on a schedule, a container with no stdin - realistically the only way `--deliver` gets used repeatedly) that wizard's `input()` call raises an **uncaught `EOFError`** instead of the clean, caught `RemarkableAPIError` every other `auth_client()` failure already produces and handles gracefully.

Verified directly against remarkapy (no mocks), with an isolated empty config file and stdin closed:

```
=== interactive=True (previous only option), no token, no stdin ===

=== REMARKABLE CLOUD ===
Visit https://my.remarkable.com/pair/app and enter the displayed code.
Verification code: confirmed: uncaught EOFError - EOF when reading a line

=== interactive=False (this PR), no token ===
clean RemarkableAPIError caught: ConfigNotFoundError No device token is configured and interactive pairing is disabled.
```

## How

- `interactive: bool = True` on `auth_client()`, forwarded to `Client(refresh_on_init=False, interactive=interactive)`. Default unchanged, so existing interactive callers see no behavior change.
- `interactive: bool = True` on `upload()`, forwarded to its own `auth_client(interactive=interactive)` call - lets a caller doing scheduled/unattended delivery opt into clean failure instead of a crash, without needing to reimplement or monkeypatch `auth_client()` themselves.
- `--no-interactive` on the standalone `upload_to_remarkable` CLI, for unattended use (cron, CI) where there's no terminal to prompt.

## Testing

- `test_auth.py` (new): parameter threading (`interactive=True`/`False` both reach the constructed `Client`), and the clean-failure path (a no-token, non-interactive `Client` produces a caught `RemarkableAPIError`, returned as `False`, not an uncaught exception).
- `test_upload.py`: `upload()`'s own default and override, confirming it reaches `auth_client()` correctly.
- Full suite: 83 passed (78 previously + 5 new).

## Merge overlap note (real, not just mechanical)

Locally merging this branch together with **#136** (adds retention pruning to `upload()`) broke 2 of #136's own tests: two of its `auth_client` mocks were still zero-arg `lambda: client`, which this PR's new `auth_client(interactive=interactive)` call breaks with a `TypeError`. Not a flaw in this PR - it already defensively updated every pre-existing `auth_client` mock it saw at the time it was written (see the diff in `test_upload.py` above) - #136 just added two more after that. Fixed on #136's side; re-verified both branches merged together now pass all tests.

### DIFF
--- a/goosepaper/auth.py
+++ b/goosepaper/auth.py
@@ -2,9 +2,17 @@ from remarkapy import Client, resolve_config_path
 from remarkapy.exceptions import RemarkableAPIError
 
 
-def auth_client():
+def auth_client(interactive: bool = True):
+    """`interactive` is forwarded to remarkapy's own `Client`, which defaults to `True` there
+    too - fine for a real terminal, where a missing device token falls into an `input()` pairing
+    wizard. In any unattended context (a cron job, a container with no stdin) that wizard's
+    `input()` call raises an uncaught `EOFError` instead of the clean, caught
+    `RemarkableAPIError` every other failure here already produces - pass `interactive=False` to
+    get that same clean handling (a caught `ConfigNotFoundError`, printed and returned as
+    `False`) instead of a crash.
+    """
     try:
-        client = Client(refresh_on_init=False)
+        client = Client(refresh_on_init=False, interactive=interactive)
         client.refresh_user_token()
         return client
     except RemarkableAPIError as err:
--- a/goosepaper/test_auth.py
+++ b/goosepaper/test_auth.py
@@ -1,0 +1,64 @@
+from remarkapy.exceptions import ConfigNotFoundError
+
+from . import auth
+
+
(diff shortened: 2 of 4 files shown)


j6k4m8 (approved):
Great! This is a much-needed feature for a lot of users I imagine :)

j6k4m8:
Once CI passes I will merge :)

PR #141, opened 2026-08-11
fix: strip readability's synthetic <html><body> wrapper unconditionally

Smengerl:
## What

Follow-up to #135 (already merged). `_make_urls_absolute` only re-serialized `body_html` via `container.decode_contents()` - which is what strips bs4/lxml's synthetic `<html><body>` wrapper around readability's `doc.summary()` output - when it actually rewrote a URL. An article whose links/images were already all absolute (common, not an edge case) hit the early `changed=False` path instead, which handed back `doc.summary()`'s raw output completely unchanged, wrapper included.

## Why this matters

A second `<html>` tag appearing mid-document is enough to confuse WeasyPrint's HTML parser into silently dropping everything from that point until it resyncs - no exception, no log line.

## Evidence

Verified live against a real 111-story daily edition (26 RSS feeds): 64 leaked `<html>` wrappers found scattered through the assembled document, several of them sitting immediately before a story that had vanished entirely from the rendered PDF.

## Fix

Always re-serialize via `container.decode_contents()`, regardless of whether anything needed absolutizing.

## Testing

- New regression test: an already-fully-absolute body still has its `<html><body>` wrapper stripped.
- Updated the three existing tests that asserted exact output equality with the raw (non-re-serialized) input string - re-serializing unconditionally now also normalizes markup (e.g. `` -> self-closing ``), which is intentional and harmless (WeasyPrint/browsers treat both identically), not a behavior regression.
- Full test suite passes (98/98).

### DIFF
--- a/goosepaper/storyprovider/rss.py
+++ b/goosepaper/storyprovider/rss.py
@@ -197,7 +197,6 @@ def _make_urls_absolute(body_html: str, base_url: str) -> str:
 
     soup = bs4.BeautifulSoup(body_html, "lxml")
     container = soup.body or soup
-    changed = False
     for tag_name, attr in (("img", "src"), ("source", "src"), ("a", "href")):
         for node in container.find_all(tag_name):
             value = node.get(attr)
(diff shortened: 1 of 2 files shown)


goosepaper/storyprovider/rss.py
```diff
@@ -209,9 +208,20 @@ def _make_urls_absolute(body_html: str, base_url: str) -> str:
...
+    # optimization to skip. bs4's lxml parser wraps whatever it's given in a synthetic
+    # <html><body> (readability's own doc.summary() output already looks exactly like a full
+    # document, so lxml has no reason to suspect it's a fragment); container.decode_contents()
+    # is what strips that wrapper back off. Returning the untouched input on a "nothing to
+    # change" run - as the original version of this function did - let that synthetic
+    # <html><body> leak straight into the final newspaper's body_html verbatim whenever an
+    # article happened to have zero relative URLs (i.e. every link/image on the page was already
+    # absolute - common, not an edge case). A second <html> tag appearing mid-document is enough
+    # to confuse WeasyPrint's parser into silently dropping everything from that point until it
+    # resyncs - verified live: found 64 such leaked wrappers across one real 111-story edition,
```
j6k4m8: don't need "verified live" text in this codebase

goosepaper/storyprovider/test_rss.py
```diff
@@ -374,11 +376,33 @@ def test_does_not_leak_a_synthetic_body_wrapper(self):
...
+    def test_strips_a_synthetic_html_body_wrapper_even_with_nothing_to_absolutize(self):
+        """Regression test for the actual production bug: bs4's lxml parser wraps whatever it's
+        given in a synthetic <html><body> (readability's doc.summary() output already looks like
+        a full document, so lxml has no reason to treat it as a fragment). Stripping that wrapper
+        via decode_contents() must happen unconditionally - the original version of this function
+        only re-serialized when it actually rewrote a URL, so an article whose links/images were
+        already all absolute (common, not an edge case) leaked the wrapper straight into the
+        newspaper's assembled HTML verbatim. A second <html> tag appearing mid-document is enough
+        to confuse WeasyPrint's parser into silently dropping everything after it until it
+        resyncs - verified live: found 64 such leaks across one real edition, several immediately
```
j6k4m8: dont need verification text in the codebase

goosepaper/storyprovider/rss.py
Smengerl: Fair point! Fixed

goosepaper/storyprovider/test_rss.py
Smengerl: Same as above. Fixed with latest commit. Found same pattern in pr133 and fixed that as well

PR #142, opened 2026-08-11
fix: re-encode RSS-embedded article images before rendering, like comics

Smengerl:
## What

RSS article images are embedded exactly as the source served them - a remote `` that WeasyPrint fetches and decodes itself while rendering the PDF, with zero control over what it gets. That reproduces a known WeasyPrint failure mode (already diagnosed and fixed for comic strips elsewhere, see #133): certain source images - oversized, wrong color mode, an unsupported format - make it silently drop the *entire* story, not just the image. No exception, no log line.

## Evidence

Verified against a real daily edition (110 stories, 26 feeds): 28 of 53 embedded images failed to make it into the rendered PDF. Every failing case traced back to the source image itself:
- multi-megapixel photos (up to 4000px, 500KB-1.4MB)
- a palette-mode PNG encoding photo-like content far less efficiently than JPEG
- WebP files - a format WeasyPrint's image backend can't decode at all, independent of size

## Fix

`_inline_remote_images` fetches each `` found in a story's assembled `body_html`, decodes and re-encodes it through Pillow (bound dimensions, normalize color mode, composite transparency onto white, always emit JPEG), and inlines the result as a `data:` URI instead of the original remote link. An image that fails to download or decode is left as its original remote link rather than aborting the whole story.

The re-encode step itself (`storyprovider/imageutil.py`) is a small shared module, not RSS-specific - see the note below.

## Testing

- New `test_imageutil.py`: direct tests of the re-encode step (oversized downscaling, CMYK->RGB, transparency composited onto white, grayscale preserved, JPEG quality honored).
- New `TestInlineRemoteImages` in `test_rss.py`: wiring tests (fetches the right URLs, skips `data:`/non-http images, tolerates a failing download without losing the rest of the story or the other images in it, end-to-end through `get_stories()`).
- Full test suite passes (109/109).

## Note: shares `imageutil.py` with #133

Comic strip images (#133) need the identical re-encode step, for the identical reason, and previously had their own separate bespoke copy of this logic. Both PRs now use the same `storyprovider/imageutil.py` - each carries its own copy of the file so neither depends on the other merging first; whichever merges second will see, at worst, a trivial identical-content conflict on that one file.

## Honesty note on scope

This fix makes image *encoding* itself reliable (0 failures out of 53 in the verification run, one intentional exception: a source SVG, which isn't a raster format Pillow can decode, left as its original link). It measurably reduces a real daily edition's total rendered size (~31% smaller in testing) and removes per-image failure as a variable. It does **not** by itself guarantee every story in a very large, image-heavy edition survives - that also depends on the document's total combined weight, a separate, document-wide WeasyPrint behavior not addressed here.

### DIFF
--- a/goosepaper/storyprovider/imageutil.py
+++ b/goosepaper/storyprovider/imageutil.py
@@ -1,0 +1,49 @@
+"""Shared image re-encoding for any externally-fetched image embedded into a goosepaper PDF.
+
+Embedding a source image unmodified - at whatever resolution, color mode/metadata, and format
+the source happens to serve - is a known way to make WeasyPrint's PDF image embedding silently
+drop content with no exception and no log line: a gap where an image (or, once combined with the
+size/weight of everything else already in a full newspaper, sometimes an entire story) should
+have been. Three contributing factors have been identified in practice: (1) some sources serve
+images at print resolution (2800px+ wide) with no smaller variant requested; (2) even at a
+source's own default resolution, a lossless PNG re-encode of photo-like or gradient-heavy content
+is itself several times larger than the same content as JPEG; (3) some sources ship CMYK-mode
+JPEGs with large embedded Photoshop/ICC metadata blocks, or formats (e.g. WebP) that WeasyPrint's
+image backend cannot decode at all. Re-encoding through Pillow first - bounding pixel dimensions,
+normalizing color mode, and always emitting JPEG - keeps every embedded image in the same
+reasonable, predictable size/format range regardless of what the source happens to serve on a
+given day.
+"""
+
+import base64
+import io
+
+from PIL import Image
+
+
+def reencode_image_as_data_uri(image_bytes: bytes, max_dimension: int, quality: int = 90) -> str:
+    """Decodes and re-encodes a fetched image as a clean, size-capped JPEG `data:` URI.
+
+    Bounds pixel dimensions to `max_dimension` on the long edge, normalizes color mode (e.g.
+    CMYK -> RGB), and composites any transparency onto white rather than leaving whatever RGB
+    value was stored underneath a transparent pixel (Pillow's plain `convert("RGB")` does not
+    composite - it just drops the alpha channel and keeps whatever was there, which can leave
+    visible phantom colors/edges where transparency was meant to show through).
+    """
+    image = Image.open(io.BytesIO(image_bytes))
+    if image.mode not in ("RGB", "L"):
+        has_transparency = (
+            image.mode in ("RGBA", "LA", "PA") or "transparency" in image.info
+        )
+        if has_transparency:
+            background = Image.new("RGB", image.size, (255, 255, 255))
+            rgba_image = image.convert("RGBA")
+            background.paste(rgba_image, mask=rgba_image.split()[-1])
+            image = background
+        else:
+            image = image.convert("RGB")
+    if max(image.size) > max_dimension:
+        image.thumbnail((max_dimension, max_dimension), Image.LANCZOS)
+    jpeg_buffer = io.BytesIO()
+    image.save(jpeg_buffer, format="JPEG", quality=quality)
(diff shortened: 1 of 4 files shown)


jpco:
Why have this image-prefetching behavior happen within story providers at all, rather than as part of rendering code, like `_render_html_document()`?  It seems that this kind of image optimization could benefit _all_ images, not just those from any particular source.  Also, `_render_html_document()` has both `layout` and `page_profile`, which could help inform a per-device maximum image width and height.

Context: See #143 - I'm interested in making images work with in epubs, which means packing all of them locally within the epub file, for a device (the Xteink X4) which is very low-power and has a relatively tiny screen.

Smengerl:
@jpco Good point, and it turned out to hold up - I spiked it on a separate branch to check before suggesting it here.

## What changed

Moved the re-encode step out of `RSSFeedStoryProvider`/`DailyComicStoryProvider` entirely and into `Goosepaper._render_html_document()` (and `to_epub()`, see below). A new `_image_max_dimension()` derives the pixel cap from the actual page geometry (`page_profile` content width / `effective_columns` * a target DPI) instead of the flat `1200` both providers guessed at independently. `_inline_story_images()` then walks every story's `body_html` - regardless of which provider produced it - fetching/decoding/normalizing any remote or already-inlined image it finds.

Result: this now protects *every* story provider uniformly, not just RSS and comics. I checked directly which of the other built-in providers can even emit an `` today - Bluesky/Reddit only ever extract plain text fields, Mastodon's attached media rides in a separate `<media:content>` RSS element none of them read - so in practice this mainly extends coverage to Readwise Reader (`body_source="html"` preserves real `` tags from saved articles), plus whatever future providers add image support without having to remember to wire up their own re-encode step.

## Numbers

Same daily edition (26 RSS feeds + XKCD + puzzles + weather/wikipedia, `page_profile: paper_pro`, 2-column), current per-provider image handling vs. this branch, generated back to back:

| | current (#142/#133) | this branch |
|---|---|---|
| stories | 101 | 101 |
| PDF pages | 75 | 75 |
| images successfully inlined | 40/44 | 40/44 |
| max embedded pixel dimension | 1200 (flat guess) | 656 (derived from `paper_pro`/2-col) |
| total embedded image payload | 2.78 MB | 1.68 MB (-40%) |

Identical reliability (same story/page/image counts), meaningfully smaller output, because images are no longer sized for a guessed "generic newspaper column" but for the actual profile/layout being rendered.

## Re: #143

The fetch+resize now lives in one place (`_inline_story_images()`) that's a natural spot to eventually take a target-device parameter for epub-specific sizing. I didn't build that - `to_epub()` currently just uses a flat fallback constant, since there's no `page_profile`/`layout` concept for a reflowable epub - but epub output does now get *some* processing where before this branch it got none at all (RSS images were remote links, comic images untouched at whatever resolution/format the source served). Real per-device sizing for #143 would still need a device parameter threaded through `to_epub()`.

## Tradeoffs I ran into (documented, not hidden)

- A story whose image fails to fetch/decode now fails soft (keeps the original unprocessed link) rather than dropping the story - matches the behavior RSS images already had, but comic strips previously had a stronger "raise and drop cleanly" guarantee at the provider level that's now gone. Open question whether that's the right tradeoff or worth restoring at the render-step level.
- Every embedded image in a document currently gets the same size cap regardless of placement (an ear/sidebar image gets the same cap as a main-column image, even though it renders narrower) - not fixed, would need `styles.py` changes to do properly.

Opened as #145, based on this repo's `master` merged with #142/#133 (rebasing onto whichever lands first once both are in).

Smengerl:
Closing in favor of #145, which supersedes this: #145's branch history includes this PR's own commits (`fix/rss-image-embedding` is an ancestor of `feature/render-time-image-sizing`) and goes further - moving image re-encoding out of individual story providers entirely into a single render-time pass shared by every provider, not just RSS. Everything this PR set out to fix is covered there.

Issue #143, opened 2026-08-13
Improved epub rendering?


jpco:
I'm interested in the epub rendering in Goosepaper in order to make it work for the Xteink X4.  Given the main use case is to generate PDFs for the reMarkable, would PRs for this slightly off-label use be welcome?

For example: I've hacked up a patch to pack images into the generated epubs, since ebooklib doesn't do it by default like weasyprint does.  Annoyingly, this does require requesting, processing, and packing the images individually, so it adds a few imports from the standard library as well as some 50 lines of code.

j6k4m8:
that would be _fantastic_, I would love to generally support more tablets but I don't have an Xteink /Boox / etc to test on. Would LOVE to get this working together :) I also think epub may be a better format for the remarkable anyway, unsure...

Woohoo! So glad you reached out :)

j6k4m8:
btw there is some work on fetching / embedding images in a current PR:
https://github.com/j6k4m8/goosepaper/pull/142/changes#diff-8bb9639aa995a02d753b06fd1688f8ecf160b1b1769e81af86c0f21959a0e010R230

I can't promise any timeline on that PR since I think it has merge conflicts right now but it's worth keeping in mind!

Smengerl:
Following up on the #142 pointer — that PR is superseded by #145, which moved image fetching/re-encoding out of individual story providers into one central pass shared by every provider.

Relevant here: `to_epub()` now calls that central pass too (it never processed images before). So the fetch/decode/resize half of your patch is already handled centrally, epub included.

Not done yet: images land as inline base64 `data:` URIs in the XHTML, not as packed `EpubItem` resources. So "packing the images individually" is still the missing piece — but should now be a much smaller patch (decode the already-inlined `data:` URIs, add each as its own `EpubItem`, rewrite `src`) rather than a full fetch+process+pack implementation.

#145 isn't merged yet but should hopefully land soon.

jpco:
Nice, that's great.  I had hoped that data URIs would work without further messing around, but it does appear that the X4 with Crosspoint (along with other e-readers) doesn't handle them at all, so packing the images separately is probably a necessity.

[...]

PR #145, opened 2026-08-15
refactor: move image re-encoding from story providers to render time

Smengerl:
## What

Follows a suggestion from @jpco on #142: image size/format normalization belongs in rendering code, which knows the actual `page_profile`/`layout` being rendered, rather than being opt-in per story provider with a guessed size constant.

Moves the image re-encode step out of `RSSFeedStoryProvider` and `DailyComicStoryProvider` (added in #142/#133) and into `Goosepaper` itself:

- New `_image_max_dimension()` derives the embedded-image pixel cap from the actual page geometry (`page_profile`'s content width / `effective_columns` * a target DPI), instead of the flat `1200` both providers guessed at independently.
- New `_inline_story_images()` walks a story's `body_html` and fetches/decodes/normalizes any `` it finds - remote `http(s)://` or an already-inlined `data:` URI. `_inline_all_story_images()` runs it over every story, isolating one story's failure from the rest.
- `_render_html_document()` (used by `to_html()`/`to_pdf()`) and `to_epub()` both call it now. `to_epub()` previously got no image protection at all from either #142 or #133 - it now does too, sized with a flat fallback constant since there's no `page_profile`/`layout` concept for a reflowable epub.
- `RSSFeedStoryProvider` no longer fetches/re-encodes images - leaves `` exactly as the source article served it.
- `DailyComicStoryProvider` no longer fetches the strip image at all - only resolves which URL is the actual strip and leaves it as a remote link, same as RSS. `image_url` is escaped before embedding (a source page's own `src` attribute value, already entity-decoded by lxml, is otherwise attacker-controlled markup).

Net effect: every story provider gets the same image protection uniformly (previously only RSS and comics were covered, and each had its own copy of the size-capping logic), and images are sized for the page they're actually rendered on instead of a guess.

## Evidence

Same daily edition (26 RSS feeds + XKCD + puzzles + weather/wikipedia, `page_profile: paper_pro`, 2-column), current per-provider image handling vs. this branch, generated back to back:

| | current (#142/#133) | this branch |
|---|---|---|
| stories | 101 | 101 |
| PDF pages | 75 | 75 |
| images successfully inlined | 40/44 | 40/44 |
| max embedded pixel dimension | 1200 (flat guess) | 656 (derived from `paper_pro`/2-col) |
| total embedded image payload | 2.78 MB | 1.68 MB (-40%) |

Identical reliability (same story/page/image counts) at meaningfully lower output size.

## Testing

- `_image_max_dimension()`: column count, margins, mm vs. in units, the fallback-on-unparseable-size path.
- `_inline_story_images()`: remote fetch, already-inlined `data:` URI, HTTP error response, non-image response, relative/missing `src`, a leading `<style>` block surviving inlining (the shape `comic.py`'s CSS + strip `<div>` has - lxml relocates a body-level `<style>` into an implied `<head>` that a naive re-serialize would drop).
- End-to-end: a real `ReadwiseReaderStoryProvider` story (the one built-in provider besides RSS/comics whose real code path can carry an `` today) through a real `Goosepaper.to_html()`.
- `to_epub()` actually inlining images, and isolating one story's image failure from the rest of the render.

Full suite passes (144 tests); `flake8` clean.

## Depends on

Built on top of #142 and #133's `imageutil.py` and the per-provider embedding they added - this branch is based on `master` merged with both of those branches, and would need to land after both are merged upstream (rebased onto whatever their final merged shape looks like).

## Known limitations (not hidden in the diff)

- A story whose image fails to fetch/decode fails soft (keeps the original, unprocessed link) rather than the story being dropped - matches the behavior RSS images already had; comic strips previously had a stronger "raise and drop cleanly" guarantee at the provider level that this removes.
- Every image in a document gets the same size cap regardless of placement - an ear/sidebar image gets the same cap as a main-column image even though it renders narrower. Would need `styles.py` changes to do properly.
- `to_epub()`'s fallback size is a flat constant, not per-device - real support for #143's use case (packing appropriately-sized images for a specific low-power e-reader) would still need a target-device parameter threaded through.


Smengerl:
Added a small follow-up fix (f6a68e5): an image that fails to inline is now removed from the story instead of being left as a dead `http(s)://` link. Leaving it in place meant WeasyPrint's own image loader tried (and failed) to fetch the exact same URL again at render time - same failure, logged twice, plus a wasted request.

Found and verified against a real case in production: cst.cam.ac.uk's own article markup has a duplicated `` tag whose `src` contains another tag's raw HTML pasted into it (their bug, not ours) - always 404s. Before: two log lines for one failure. After: one, and the PDF still renders correctly (the page's other, correctly-formed `` for the same photo still embeds fine). 145/145 tests pass.

